PR #1, opened 2019-01-03
Adding the reduced fileset from the cookie cutter edx publisher app

staubina:


documentation/.travis.yml.md
```diff
@@ -0,0 +1,77 @@
+# Travis Configuration
```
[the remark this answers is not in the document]
staubina: Had this locally, should have pushed it.

staubina:
Have the .idea in gitignore locally.  pushing soon.

.npmignore
```diff
@@ -0,0 +1,13 @@
+.eslintignore
```
jlajoie: Do we need an NPM ignore? I'm under the impression we won't be publishing this project to NPM?

Makefile
```diff
@@ -0,0 +1,32 @@
+shell: ## run a shell on the cookie-cutter container
+	docker exec -it edx.fecc /bin/bash
```
jlajoie: Will this need to change to edx.pubfe?

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+ARG PORT=80
+ENV PORT $PORT
+EXPOSE $PORT 18400
+
+WORKDIR /edx
+# Install app dependencies
+# A wildcard is used to ensure both package.json AND package-lock.json are copied
+# where available (npm@5+)
+COPY package*.json ./
```
jlajoie: I think I talked to you about this in person but just capturing it here, is there a reason we're copying this to the /edx dir rather than /edx/app?

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+version: "2"
```
jlajoie: We use 2.1 for `devstack`, should this be upgraded to match it? Though I think compose is up to 3 now.

Dockerfile
staubina: I am not sure, we should ask around.  Maybe @adamstankiewicz or @georgebabey might be able to provide insight.

package.json
```diff
@@ -0,0 +1,96 @@
...
+    "email-prop-type": "^1.1.5",
+    "font-awesome": "^4.7.0",
+    "history": "^4.7.2",
+    "prop-types": "^15.5.10",
+    "react": "^16.2.0",
+    "react-dom": "^16.2.0",
+    "react-redux": "^5.0.7",
+    "react-router": "^4.2.0",
+    "react-router-dom": "^4.2.2",
+    "react-router-redux": "^5.0.0-alpha.9",
```
jlajoie: Do we want to be locked into an alpha version for this?

docker-compose.yml
staubina: If we need something from 3 I am ok to upgrade to it in the future.  I will set it to 2.1 for now if that will work.

.npmignore
staubina: I am not sure. I can remove it and we can add it back later if we find it is needed.

Dockerfile
staubina: Tried changing this locally and all still seems to work. I will push an update.

georgebabey: Makes sense to me to put into /edx/app

We have not been using docker for development but instead have been using our local machines so something we've never ran into

PR #4, opened 2019-01-08
Documenting the decision for long term configuration needs.

staubina:


decisions/0001-long-term-configuration.rst
```diff
@@ -0,0 +1,32 @@
...
+
+Status
+------
+
+Initial
+
+Context
+-------
+
+Overtime configuration will need to be added to the application.  This needs to be separated between local and
```
mikix: grammar nit: over time is two words in this case, I think.

PR #19, opened 2019-01-17
Header and footer. Woo (exclamation point)

Dillon-Dumesnil:


src/App.scss
```diff
@@ -1,2 +1,4 @@
 @import "~@edx/edx-bootstrap/sass/edx/theme";
 @import "~bootstrap/scss/bootstrap";
+
+@import "~@edx/frontend-component-footer/src/lib/scss/site-footer";
```
jlajoie: nit: Why blank line

src/index.jsx
```diff
@@ -30,6 +31,24 @@ const App = () => (
...
+          supportUrl={process.env.SUPPORT_URL}
+          contactUrl={process.env.CONTACT_URL}
+          openSourceUrl={process.env.OPEN_SOURCE_URL}
+          termsOfServiceUrl={process.env.TERMS_OF_SERVICE_URL}
+          privacyPolicyUrl={process.env.PRIVACY_POLICY_URL}
+          facebookUrl={process.env.FACEBOOK_URL}
+          twitterUrl={process.env.TWITTER_URL}
+          youTubeUrl={process.env.YOU_TUBE_URL}
+          linkedInUrl={process.env.LINKED_IN_URL}
+          googlePlusUrl={process.env.GOOGLE_PLUS_URL}
```
jlajoie: Should we leave a not that this is going away soon?

src/App.scss
Dillon-Dumesnil: Separates by app scss we are pulling in. If we end up using more apps, the lines will help

src/index.jsx
Dillon-Dumesnil: :shrug: Nah. We'll just remove it when it's time to

PR #27, opened 2019-01-18
Aj/disco 479 create course

staubina:
WIP PR to provide context on my current work.


src/components/CreateCoursePage/CreateCoursePage.scss
```diff
@@ -0,0 +1,25 @@
+@import "~bootstrap/scss/_functions.scss";
+@import "~bootstrap/scss/_variables.scss";
+
+.create-course-form {
+  label {
+    font-weight: 600;
+  }
```
mikix: Maybe `@extend .font-weight-bold`, to standardize weights. (or find a way to add that class to the parent element directly in our React code to avoid the custom sass?)

src/components/CreateCoursePage/CreateCoursePage.scss
```diff
@@ -0,0 +1,25 @@
...
+
+  .required {
+    color: $danger;
+    margin-left: 4px;
+  }
```
mikix: Maybe the following?  Makes elements more themable, in case a theme wants to make text bold in danger case or whatever. Could also do this directly in React, but looks like we use this a few places.

```
@extend .ml-1 // (or 2? not sure right size)
@extend .text-danger
```

src/components/TableComponent/__snapshots__/TableComponent.test.jsx.snap
```diff
@@ -1150,6 +1150,7 @@ ShallowWrapper {
       },
     },
   },
+  Symbol(enzyme.__childContext__): null,
```
staubina: I am not sure why these snapshots are in here.  Looking into it before merging PR.

src/data/constants/index.js
```diff
@@ -0,0 +1,3 @@
+export const VERIFIED_TRACK = { name: 'Verified', key: 'verified' };
+export const AUDIT_TRACK = { name: 'Audit', key: 'audit' };
+export const PROFESSIONAL_TRACK = { name: 'Professional', key: 'professional' };
```
mikix: When are we gonna start marking strings for translation?

src/data/reducers/courseInfo.js
```diff
@@ -31,6 +35,27 @@ function courseInfo(state = initialState, action) {
         isFetching: true,
         error: null,
       });
+    case CREATE_COURSE:
+      return Object.assign({}, state, {
+        data: {},
+        courseCreated: false,
+        isFetching: false,
+        error: null,
+      });
```
mikix: 1) courseCreated... It's used as a signal to switch pages, yeah? I guess that's the right pattern? Feels weird to me, but alright.

2) You're setting isFetching but don't care about isFetching. You can just leave it out, and it won't be touched. (the current state will come in via the `state` argument to `Object.assign` and stay the same if you don't override it)

src/data/actions/courseInfo.js
```diff
@@ -20,6 +23,20 @@ export function requestCourseInfo(id) {
   return { type: REQUEST_COURSE_INFO, id };
 }
 
+export function createNewCourse(courseData) {
+  return {
+    type: CREATE_COURSE, courseData,
+  };
+}
```
mikix: I don't care, really. But this is a different formatting than others.

src/data/actions/courseInfo.js
```diff
@@ -50,3 +67,18 @@ export function fetchCourseInfo(id) {
...
+    return DiscoveryDataApiService.createCourse(courseData)
+      .then((response) => {
+        const course = response.data;
+        dispatch(courseCreateSuccess(course));
+      })
+      .catch((error) => {
+        dispatch(courseCreateFail(`Course create failed, please try again or contact support. Error( ${error.response.data} )`));
+      });
+  };
+}
```
mikix: Guh! All this boilerplate for the same idea. "Do a request, handle request success/fail" - the fedx recommendation was just writing a factory class for these. We should look at that at some point. For now, it's just easier to copy and paste. But really.

src/containers/CreateCourse/index.jsx
```diff
@@ -0,0 +1,26 @@
...
+function mapDispatchToProps(dispatch) {
+  return {
+    fetchOrganizations: () => {
+      dispatch(fetchOrganizations());
+    },
+    createCourse: (courseData) => {
+      dispatch(createCourse(courseData));
+    },
+  };
+}
```
mikix: I believe you can just write this as:

```
const mapDispatchToProps = {
  createCourse,
  fetchOrganizations,
};
```

And redux will do this boilerplate for you.

src/containers/CreateCourse/CreateCourse.test.jsx
```diff
@@ -0,0 +1,85 @@
...
+    testState.publisherUserInfo.isFetching = false;
+    testState.publisherUserInfo.error = errorMessage;
+    wrapper = createWrapper(testState).dive();
+    expect(wrapper.find('#error')).toHaveLength(1);
+    expect(wrapper.find('#error').props().dialog).toEqual(errorMessage);
+  });
+
+  it('Submits the form with correct data', () => {
+    const testState = initialState;
+    // const handleSubmit = jest.fn();
```
mikix: Looks leftover.

src/containers/CreateCourse/CreateCourse.test.jsx
```diff
@@ -0,0 +1,85 @@
...
+const createWrapper = (state) => {
+  store = mockStore(state);
+  mockOnSubmit = jest.fn();
+  return shallow(<CreateCoursePage store={store} createCourse={mockOnSubmit} />);
+};
+
+
+describe('CreatCourse View', () => {
+  beforeEach(() => {
+  });
```
mikix: Could drop.

src/components/TableComponent/__snapshots__/TableComponent.test.jsx.snap
mikix: If you pulled in enzyme 3.9.0? That might explain it.

src/components/CreateCoursePage/index.jsx
```diff
@@ -0,0 +1,149 @@
...
+
+    if (publisherUserInfo.isFetching) {
+      return (
+        <div className="mx-auto text-center">
+          <Icon
+            id="spinner"
+            className={['fa', 'fa-circle-o-notch', 'fa-spin', 'fa-3x', 'fa-fw']}
+          />
+        </div>
+      );
```
mikix: We might want to pull this spinner construction into a component.

src/components/CreateCoursePage/index.jsx
```diff
@@ -0,0 +1,149 @@
...
+  showErrorStatus(error) {
+    return (
+      <StatusAlert
+        id="error"
+        alertType="danger"
+        dismissible={false}
+        open
+        dialog={error}
+        className={['text-center', 'mx-auto', 'w-50']}
+      />
```
mikix: We might want to pull this specific style of status alert into a simple component that just takes a string.

src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -0,0 +1,147 @@
...
+            <form onSubmit={handleSubmit}>
+              <Field
+                name="courseOrg"
+                component={RenderSelectField}
+                options={this.processOrganizations(organizations)}
+                label={
+                  <React.Fragment>
+                    Organization
+                    <span className="required">*</span>
+                  </React.Fragment>
```
mikix: Looking up what react.fragment does, I saw that you can also just write it like:

```
                  <>
                    Organization
                    <span className="required">*</span>
                  </>
```

But that feels... weird. And if I saw that, I wouldn't have known what to search for to see what it was.

src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -0,0 +1,147 @@
...
+              />
+              <Field
+                name="courseTitle"
+                component={RenderInputTextField}
+                type="text"
+                label={
+                  <React.Fragment>
+                    Course title
+                    <span className="required">*</span>
+                  </React.Fragment>
```
mikix: This feels like maybe it could be pulled into a component like "FormLabel" that takes a required parameter and adds the necessary asterisk or styling?

Does the "required" styling come from bootstrap?

src/components/CreateCoursePage/index.jsx
staubina: I agree, should I do that in this PR you think?

src/components/CreateCoursePage/CreateCourseForm.jsx
staubina: I agree that it looks like something I would delete.  The `<React.Fragment>` seems more clear to me.

src/components/CreateCoursePage/CreateCourseForm.jsx
staubina: No, this was a style that was removed when copying the code over.  I think we can remove it and add style later, when we decide on the over all style of the pages, do you agree?

src/data/constants/index.js
staubina: That is a good question.  I will add that to the list of things to plan for.

src/data/reducers/courseInfo.js
staubina: For courseCreated, I would be interested to hear about another approach you might have in mind. I was using it to signal the course was created successfully.

src/containers/CreateCourse/index.jsx
staubina: 👍

src/components/CreateCoursePage/index.jsx
staubina: Done

src/components/CreateCoursePage/index.jsx
staubina: I am using the one @jlajoie already added now.

src/data/reducers/courseInfo.js
staubina: I was able to update this and make it simpler I think.  Take a look at the PR now.

staubina:
@mikix I made a number of updates and refactors after reading your comments.  Please take a look as it ended up in a pretty substantial refactor.

src/data/actions/courseInfo.js
Dillon-Dumesnil: I would just change them all the match the constant you're using.
createCourseInfo (or something like that. Mine is editCourseInfo)
createCourseSuccess
createCourseFail

src/components/CreateCoursePage/CreateCourseForm.jsx
mikix: Yeah, I don't have a strong opinion on whether we style it now or later. If you wanted red now without custom sass, you could replace `required` with `text-danger` which comes from bootstrap and gives a red color.

I do still like the idea of componentizing this. The color, whether the title is bolded, whether we use an asterisk, etc. are going to be the same across the app. Easier to pull it into its own thing.

src/data/reducers/courseInfo.js
mikix: Oh nice, moving the push to the fetch callback does simplify the component caring about that state. It is now maybe a bit presumptiuous of the fetch callback to assume the page wants to be switched? But we only call it from the one place for now anyway.

I still think it's unnecessary to set isFetching, since that's not part of the create-course action flow. But it isn't harmful.

src/components/CreateCoursePage/index.jsx
```diff
@@ -0,0 +1,137 @@
...
+          <title>Create a New Course</title>
+        </Helmet>
+
+        <div className="container-fluid">
+          <div className="row justify-content-md-center my-3 ">
+            <div className="col-6">
+              { publisherUserInfo.isFetching && <LoadingSpinner /> }
+              { publisherUserInfo.error && this.showErrorStatus(publisherUserInfo.error) }
+              { !publisherUserInfo.isFetching && this.renderForm() }
+            </div>
```
mikix: Looks nice and clean! :)

PR #29, opened 2019-01-22
Fix nested path URLs not finding webfonts

mikix:
The <base> change is not strictly required. I think firefox was able to determine what we were trying to say regardless. But figured this is more compliant.


mikix:
Yes we should. I was going to prep the publicPath change

Dillon-Dumesnil:
Yeah, the only difference is the publicPath in webpack. The index.html file now reflects what is in the cookiecutter.

PR #30, opened 2019-01-23
DISCO-471: Adds in Course Dashboard display with Pagination

jlajoie:
Pagination and Table interactions duplicated from edx-portal.


src/components/StatusAlert/StatusAlert.scss
```diff
@@ -0,0 +1,12 @@
+.alert {
+  .icon {
+    .fa  {
+      font-size: 1.5rem;
+    }
+  }
+
+  .title {
+    font-weight: 600;
+    display: block;
```
AlasdairSwan: Alternatively you could use the util class ```d-block``` at the element level.

src/components/TableComponent/TableComponent.scss
```diff
@@ -0,0 +1,11 @@
+.table {
+  font-size: 0.875rem;
+
+  th.sortable > button {
+    padding-left: 0;
+    font-size: 0.875rem;
+    font-weight: 600;
```
AlasdairSwan: If you hook up the linter this will throw an error, and suggest you should rewrite as;
```
font: {
  size: 0.875rem;
  weight: $font-weight-bold;
}
```

src/components/TableComponent/TableComponent.scss
```diff
@@ -0,0 +1,11 @@
+.table {
+  font-size: 0.875rem;
+
+  th.sortable > button {
```
AlasdairSwan: Styles should be applied to classes instead of elements.

src/components/TableComponent/index.jsx
```diff
@@ -0,0 +1,182 @@
...
+      error,
+    } = this.props;
+
+    return (
+      <React.Fragment>
+        {error && this.renderErrorMessage()}
+        {loading && !data && this.renderLoadingMessage()}
+        {!loading && !error && data && data.length === 0 &&
+          this.renderEmptyDataMessage()
+        }
```
AlasdairSwan: I guess this comes down to preference but the first 3 render functions return very small jsx elements so I would advocate for just moving them directly into the render function.

src/components/TableComponent/index.jsx
```diff
@@ -0,0 +1,182 @@
...
+        {data && data.length > 0 && this.renderTableContent()}
```
AlasdairSwan: It may be worth moving ```renderTableContent``` to its own component.

config/webpack.dev.config.js
```diff
@@ -121,6 +121,7 @@ module.exports = Merge.smart(commonConfig, {
       REDDIT_URL: 'https://www.reddit.com',
       APPLE_APP_STORE_URL: 'https://www.apple.com/ios/app-store/',
       GOOGLE_PLAY_URL: 'https://play.google.com/store',
+      DATA_API_BASE_URL: 'http://localhost:18381',
```
staubina: Should this have Discoveries name in the constant? I think it will make it more clear.

src/components/TableLoadingOverlay/TableLoadingOverlay.scss
```diff
@@ -0,0 +1,19 @@
+.table-loading-overlay {
+  position: absolute;
+  z-index: 10;
+  background: #000000;
```
AlasdairSwan: Variables should be used for colours.

src/components/TableLoadingOverlay/TableLoadingOverlay.scss
```diff
@@ -0,0 +1,19 @@
...
+  width: 100%;
+}
+
+.table-loading-message {
+  position: absolute;
+  z-index: 11;
+  width: 100%;
+  height: 100%;
+
+  span {
```
AlasdairSwan: Styles should be applied to classes instead of elements.

src/components/TableLoadingOverlay/index.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import LoadingMessage from '../LoadingMessage';
+
+import './TableLoadingOverlay.scss';
+
+const TableLoadingOverlay = () => (
+  <React.Fragment>
+    <div className="table-loading-overlay" />
```
AlasdairSwan: It would be worth adding ```aria-hidden="true"``` to cover our a11y bases.

src/components/LoadingMessage/LoadingMessage.scss
```diff
@@ -0,0 +1,14 @@
+.loading {
+  background: #f2f8fb;
+  font-size: 1.125rem;
+
+  &.table-loading,
+  &.request-codes {
```
staubina: This looks like it is from edx-portal, can we remove it?

src/components/LoadingMessage/LoadingMessage.scss
```diff
@@ -0,0 +1,14 @@
...
+    height: 360px;
+  }
+
+  &.overview,
+  &.coupons {
```
staubina: This looks like it is from edx-portal, can we remove it?

config/webpack.dev.config.js
jlajoie: Yes it should, bad rebase in here!

src/components/LoadingMessage/index.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+
+import './LoadingMessage.scss';
```
AlasdairSwan: So is the CSS compiled at the component level and inserted into the DOM? Although prospectus doesn't do it, I still prefer external stylesheets (and plan on updating prospectus at some point to use that approach).

src/components/TableComponent/TableComponent.test.jsx
```diff
@@ -0,0 +1,160 @@
...
+      fetchMethod={mockFetch}
+      columns={mockColumns}
+      formatData={mockFormat}
+      tableSortable
+      paginateTable={() => true}
+      sortTable={() => true}
+      clearTable={() => true}
+      location={{ search: '?page=1&ordering=key' }}
+    />);
+    expect(component).toMatchSnapshot();
```
staubina: I am still new to snapshots, but should we also check for its presence in the DOM?  My understanding is that the snapshot will only ensure we do not change the structure of the DOM unexpectedly.

src/components/TableComponent/TableComponent.test.jsx
```diff
@@ -0,0 +1,160 @@
...
+      clearTable={() => true}
+      location={{ search: '?page=1&ordering=key' }}
+      data={mockFetch()}
+    />);
+    component.setProps({
+      location: {
+        search: '?page=1&ordering=-key',
+      },
+    });
+    expect(component).toMatchSnapshot();
```
staubina: Same question as above on all these tests.

src/containers/TableContainer/TableContainer.test.jsx
```diff
@@ -0,0 +1,39 @@
...
+        },
+      },
+    };
+    const mappedState = mapStateToProps(initialState, { id: tableId });
+    expect(mappedState.data).toEqual(initialState.table[tableId].data.results);
+    expect(mappedState.currentPage).toEqual(initialState.table[tableId].page);
+    expect(mappedState.pageCount).toEqual(1); // data.count / pageSize
+    expect(mappedState.ordering).toEqual(initialState.table[tableId].ordering);
+    expect(mappedState.loading).toEqual(initialState.table[tableId].loading);
+    expect(mappedState.error).toEqual(initialState.table[tableId].error);
```
staubina: Ok, I think I see now that you are testing for the presence of the components at this level.

src/data/actions/table.js
```diff
@@ -0,0 +1,154 @@
...
+      }
+      dispatch(paginationFailure(tableId, error));
+    });
+  }
+);
+
+const sortBy = (dataToSort, orderField) => {
+  const isFloatValue = value => !Number.isNaN(value) && !Number.isNaN(parseFloat(value));
+  const sortByOptions = (value1, value2) => {
+    let a = value1[orderField] || '';
```
staubina: nit: I am not a huge fan of 1 letter variable names, but in no way a blocker.

src/data/actions/table.js
```diff
@@ -0,0 +1,154 @@
...
+  (dispatch, getState) => {
+    const tableState = getState().table[tableId];
+    const options = {
+      ...getPageOptionsFromUrl(),
+      ordering,
+    };
+    dispatch(sortRequest(tableId, ordering));
+
+    // If we can sort client-side because we have all of the data, do that
+    if (tableState.data && tableState.data.num_pages === 1) {
```
staubina: This is a nice addition 👍

src/components/CourseTable/index.jsx
```diff
@@ -1,13 +1,47 @@
...
+    {
+      label: 'Modified',
+      key: 'modified',
+      columnSortable: true,
+    },
+  ];
+
+  const formatCourseData = courses => courses.map(course => ({
+    ...course,
+    owners: course.owners ? course.owners.map(owners => owners.name).join(', ') : '',
```
staubina: In my local data I am not seeing owner coming over with my test courses.  I need to check where it is set.  This might just be something we are not currently setting with the new Create API.

staubina: Looks like an issue with the new Course creation code.  I will move it to separate ticket.

staubina:
Tested locally with a small set of courses and it looks great.  Let me know if you have questions about my review.

PR #43, opened 2019-02-21
Added a New Course button

staubina:
Quick PR to Center the table and add a New Course Button

### DIFF
--- a/src/components/CourseTable/index.jsx
+++ b/src/components/CourseTable/index.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { Link } from 'react-router-dom';
 
 import TableContainer from '../../containers/TableContainer';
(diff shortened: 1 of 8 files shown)


src/components/CourseTable/index.jsx
```diff
@@ -33,14 +34,29 @@ const CourseTable = () => {
...
-      columns={courseTableColumns}
-      formatData={formatCourseData}
-      tableSortable
-    />
+    <div className="container">
+      <div className="row justify-content-md-center">
+        <div className="col-md-10">
+          <div className="row justify-content-md-end">
+            <Link to="/courses/new">
+              <button className="btn-outline-primary">New Course</button>
```
mikix: Probably want the "btn" class too (that I think handles some common cases like disabled state)

staubina: Oh Thanks

PR #45, opened 2019-02-21
Updates based on feedback on CourseCreate Form

staubina:


src/components/CreateCoursePage/index.jsx
```diff
@@ -7,6 +7,11 @@ import StatusAlert from '../StatusAlert';
 import LoadingSpinner from '../LoadingSpinner';
 
 class CreateCoursePage extends React.Component {
+  constructor(props) {
+    super(props);
+    this.handleCourseCreate = this.handleCourseCreate.bind(this);
```
staubina: This change along with the change to onSubmit breaks my Submit test :(. @AlasdairSwan Could you help me fix that breakage?

AlasdairSwan: What is the error? That submit is not being called?

staubina: All set now

PR #47, opened 2019-02-22
Added course link to header

staubina:


src/components/Header/index.jsx
```diff
@@ -13,12 +14,18 @@ export default class Header extends React.Component {
...
-        <header className="d-flex justify-content-left align-items-center p-3 border-bottom-blue">
-          <Hyperlink content={this.renderLogo()} destination="https://www.edx.org" />
-          <span className="badge badge-secondary beta">Alpha</span>
-          <div />
-        </header>
+      <div className="row mb-3">
+        <div className="col">
+          <header className="d-flex justify-content-left align-items-center p-3 border-bottom-blue">
+            <div className="col-md-auto">
+              <Hyperlink content={this.renderLogo()} destination="https://www.edx.org" />
```
jlajoie: This should probably reference an env var

staubina: Yah, this was a preexisting issue.  I think I will change it real quick though.

PR #48, opened 2019-02-26
Mikix/course edit

Dillon-Dumesnil:


src/components/CourseTable/index.jsx
```diff
@@ -30,7 +30,7 @@ const CourseTable = () => {
 
   const formatCourseData = courses => courses.map(course => ({
     ...course,
-    title: (<Link to={`/courses/${course.uuid}`}>{course.title}</Link>),
+    title: (<Link to={`/courses/${course.uuid}/edit/`}>{course.title}</Link>),
```
staubina: Something we should check in on as a team: Should the URL here just be /courses since we will not have a View Course?

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -0,0 +1,208 @@
...
+
+  render() {
+    const {
+      handleSubmit,
+      number,
+      price,
+      submitting,
+    } = this.props;
+
+    const options = this.getOptions();
```
staubina: Will this call the Get method for options on every render?

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -0,0 +1,208 @@
...
+              required
+            />
+            <div className="row">
+              <div id="number" className="col-12">
+                <strong>Number: *</strong>
+              </div>
+              <div className="col-12">{number}</div>
+            </div>
+            <Field
+              name="short_description"
```
staubina: Should we be using camel case on all of the field names?

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -0,0 +1,208 @@
...
+              component={RichEditor}
+              label={<strong>Short Description: *</strong>}
+              maxChars={500}
+              id="sdesc"
+            />
+            <Field
+              name="full_description"
```
staubina: Same as above.

src/components/EditCoursePage/index.jsx
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  componentDidMount() {
+    this.props.fetchCourseInfo();
+    this.props.fetchCourseOptions();
+    this.setStartedFetching();
+  }
+
+  setStartedFetching() {
+    this.setState({ startedFetching: true });
```
staubina: I am confused as to why we need to set state in the view?  

Could this be handled with the redux actions and reducers?

If the first thing we are doing is fetching shouldn't the initial state be fetching?

src/components/EditCoursePage/index.jsx
```diff
@@ -0,0 +1,213 @@
...
+          alertType="danger"
+          title="Course Edit Form failed to load: "
+          message="Course information unavailable."
+        />
+      );
+    }
+
+    const {
+      courseInfo,
+      courseInfo: {
```
staubina: Can you just map courseInfo here? Or do you have to call out all the sub-objects you are using?

src/components/EditCoursePage/index.jsx
```diff
@@ -0,0 +1,213 @@
...
+          video,
+          entitlements,
+        },
+      },
+      courseOptions,
+    } = this.props;
+
+    // If we want to keep a lot of the logic in the lower return,
+    // we have to do all this weird '&&' logic to make sure the data is there.
+    const keyParts = key && key.split(/\+|\//);
```
staubina: We should avoid special logic to try and parse the key if we can.  Do we need this key split here?

If we do we need to be cautious, any special logic to parse the Course Key and Run key concerns me.

Will this handle new and old style keys?

Also, if we have no other option and we need to do this we should move this to a method so that later if we need it again we can move it to some common tools.

src/components/EditCoursePage/index.jsx
```diff
@@ -0,0 +1,213 @@
...
+    return (
+      <React.Fragment>
+        <Helmet>
+          <title>{`Course - ${title}`}</title>
+        </Helmet>
+
+        <div className="container-fluid">
+          <div className="row justify-content-md-center my-3 ">
+            <div className="col-6">
+              { (!this.state.startedFetching || courseInfo.isFetching ||
```
staubina: Do we need the startedFetching flag?

src/components/EditCoursePage/index.jsx
```diff
@@ -0,0 +1,213 @@
...
+                courseOptions.isFetching) &&
+                <LoadingSpinner /> }
+              { courseInfo.error && (
```
staubina: Could you merge these into one status alert with a single conditional?

Maybe something like
```
error = //Combine the error messages
{ error && (
<StatusAlert
   id="error"
   alertType="danger"
   title="Course Edit Form failed to load: "
   message={courseInfo.error}
/>
)}
```
This is just sudo code of course, but it seems possible.  Unless the intent is to show multiple status alerts, which is also valuable.

src/data/reducers/courseInfo.test.js
```diff
@@ -53,11 +53,13 @@ describe('courseInfo reducer', () => {
       });
   });
 
+  /* isFetching is `true` for all of the course create tests because course create shouldn't
+  alter isFetching. Since oldState has it set to true, course create will maintain the value */
```
staubina: We should review this, It should not be true here.  I am also looking at this in my PR now.

src/components/CourseTable/index.jsx
Dillon-Dumesnil: Happy for whatever. Let me know.

src/components/EditCoursePage/EditCourseForm.jsx
Dillon-Dumesnil: Nope. This is just grabbing the data and assigning it to options. We fetch the options as part of the Edit Page before we do anything with the Form.

src/components/EditCoursePage/EditCourseForm.jsx
Dillon-Dumesnil: We use snake case because the name of the field is what is returned when the form is submitted and what we eventually send to discovery. These are written so that they match the discovery fields for updating.

src/components/EditCoursePage/index.jsx
Dillon-Dumesnil: Re question 1: we are using this state variable as an indicator for when we have actually started fetching or not. This is because isFetching is initially False and without having this state set, it will try to render the component without any data and cause it to crash.

Re question 2: It might be able to be handled with redux, but I couldn't come up with a good way to set it and thought state might be the best. 

Re question 3: I kind of disagree with that because there is a time period before we start fetching so initializing it to `true` feels inaccurate.

src/components/EditCoursePage/index.jsx
Dillon-Dumesnil: This is the preferred method of creating all of those objects. Alasdair was the one who advised this.

src/components/EditCoursePage/index.jsx
Dillon-Dumesnil: We do need something because we don't want to display the key and want to display only the course number.

It does handle the case of the key being split by `+` and by `/` so I think it handles both styles. Happy to move it to a util function if we think that makes the most sense.

src/components/EditCoursePage/index.jsx
Dillon-Dumesnil: See above comment about setting state.

src/components/EditCoursePage/index.jsx
Dillon-Dumesnil: I don't have a preference. Happy to do whichever. I haven't looked into what the error messages would look like so I'm not really sure whether it makes more sense to combine or not.

src/data/reducers/courseInfo.test.js
staubina: https://github.com/edx/publisher-frontend/pull/58/

src/components/EditCoursePage/index.jsx
staubina: I think it is ok to keep it as is. I looked at it again and it 2 error boxes will be fine.  We can adjust with experience team feedback as needed.

src/components/EditCoursePage/index.jsx
staubina: Let's at least move it to a method and then if we need it again move it to a utility.

src/components/EditCoursePage/index.jsx
staubina: It sounds like you have thought it out.  I do not feel that I need to block on this change.

Thank you for the detail.

src/components/CourseTable/index.jsx
staubina: We can postpone this change for a later date.

src/components/EditCoursePage/index.jsx
Dillon-Dumesnil: I think we should make the decision to either move it straight to a utility or keep it as is since it is only being used once right now. Moving it into a method is more lines of code if it's only being called once.

staubina: The question is, what is the likelihood we will use this again elsewhere in the code?

If you think it will not, I am fine with it staying where it is.

I am ok with it not moving, but please add a named constant variable for the regex.  Just helps with code readability over time.

src/components/EditCoursePage/index.jsx
```diff
@@ -0,0 +1,213 @@
...
+          syllabus_raw,
+          video,
+          entitlements,
+        },
+      },
+      courseOptions,
+    } = this.props;
+
+    // If we want to keep a lot of the logic in the lower return,
+    // we have to do all this weird '&&' logic to make sure the data is there.
```
staubina: We might want to avoid calling the logic "weird" in the comment, makes it seems like we should remove it.

src/components/RichEditor/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+      charCount: content.length,
+    });
+    const htmlContent = editor.getContent();
+    this.props.input.onChange(htmlContent);
+  }
+
+  render() {
+    const remainingChars = this.props.maxChars - this.state.charCount;
+    const characterLimitMessage = `Recommended character limit (including spaces) is
+      ${this.props.maxChars}. ${remainingChars} characters remaining.`;
```
AlasdairSwan: ```this.props.maxChars``` is used more than once in the render function so I would recommend destructuring it.

src/data/constants/courseOptions.js
```diff
@@ -0,0 +1,3 @@
+export const FAIL_COURSE_OPTIONS = 'FAIL_COURSE_OPTIONS';
+export const RECEIVE_COURSE_OPTIONS = 'RECEIVE_COURSE_OPTIONS';
+export const REQUEST_COURSE_OPTIONS = 'REQUEST_COURSE_OPTIONS';
```
AlasdairSwan: Why not continue the convention you updated the previous file to follow (i.e. define the constants and then export them)?

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -0,0 +1,208 @@
...
+              name="outcome"
+              component={RichEditor}
+              label={<strong>What you will learn: *</strong>}
+              maxChars={2500}
+              id="outcome"
+            />
+            <Field
+              name="subjectPrimary"
+              component={RenderSelectField}
+              label={<strong>Primary Subject: *</strong>}
```
mikix: I'm still a proponent of having a `FieldLabel` utility component that takes a required flag and a label string and does the right styling. Like maybe makes the asterisk red. Or use some fancy unicode asterisk. Or whatever we decide eventually.

It's fine to not take that on in this PR. But just noting it again for posterity.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -0,0 +1,208 @@
...
+              type="text"
+              label={<strong>About Video Link: *</strong>}
+              required
+            />
+            <Field
+              name="mode"
+              component={RenderInputTextField}
+              type="text"
+              label={<strong>Course Mode:</strong>}
+              disabled
```
mikix: This is fine for now, but we do have plans to allow adjusting mode right?

This field and the next also prefix "Course" in the label. That's maybe clearer than without it? I dunno. Just noting it.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -0,0 +1,208 @@
...
+              component={RichEditor}
+              label={<strong>Additional Information:</strong>}
+              maxChars={2500}
+              id="additional-information"
+            />
+            <Field
+              name="syllabus_raw"
+              component={RichEditor}
+              label={<strong>Syllabus:</strong>}
+              maxChars={500}
```
mikix: Does Field just pass on all unknown fields to the component? Neat.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -0,0 +1,208 @@
...
+              type="text"
+              label={<strong>Course Mode:</strong>}
+              disabled
+            />
+            {price && <Field
+              name="price"
+              component={RenderInputTextField}
+              type="number"
+              label={<strong>Course Price: *</strong>}
+            />}
```
mikix: This should use some sort of "price" formatting. So "2.50" shows instead of "2.5" - is that easy?

src/components/EditCoursePage/index.jsx
Dillon-Dumesnil: I'm just going to add it into utils.js just in case and then it can be used anywhere. Do you still think the constant variable is necessary once we have
```
function getCourseNumber(courseKeyFragment) {
  const keyParts = courseKeyFragment.split(/\+|\//);
  return keyParts[keyParts.length - 1];
}
```

src/components/EditCoursePage/index.jsx
```diff
@@ -0,0 +1,213 @@
...
+                  message={courseOptions.error}
+                />
+              )}
+              { (this.state.startedFetching && !courseInfo.isFetching &&
+                !courseOptions.isFetching) && (
+                <div>
+                  <h2>Course</h2>
+                  <hr />
+                  <div className="col">
+                    <EditCourseForm
```
mikix: The alignment of this component inside the broader component isn't quite right. The left and right edges don't line up with the "Course" title and the "Save Course" button. Doesn't need to be done here - we can beautfy later. But just mentioning.

src/components/EditCoursePage/EditCourseForm.jsx
bbaker6225: Some of the logic from [prospectus formatting](https://github.com/edx/prospectus/blob/master/src/utils/formatting.js#L5-L19) might be helpful here.  The tests [here](https://github.com/edx/prospectus/blob/master/src/utils/formatting.test.js) give some examples of how it works

src/data/constants/courseOptions.js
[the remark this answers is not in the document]
Dillon-Dumesnil: I agree and dislike the two different orderings. Let's chat at standup on Monday, choose a convention, and I'll just update them all so we can resolve this once and for all.

src/utils.js
```diff
@@ -45,7 +46,19 @@ const getPageOptionsFromUrl = () => {
...
+function jsonDeepCopy(src) {
+  return JSON.parse(JSON.stringify(src));
+}
+
+function getCourseNumber(courseKeyFragment) {
+  const COURSE_KEY_FRAGMENT_REGEX = /\+|\//;
+  const keyParts = courseKeyFragment.split(COURSE_KEY_FRAGMENT_REGEX);
+  return keyParts[keyParts.length - 1];
+}
+
```
AlasdairSwan: Test for these would be nice.

PR #55, opened 2019-03-06
Add staff grid component

bbaker6225:
Early WIP PR for the staff grid. Let me know if there are any early concerns, or if we have more details about what info we want to show in the regular view
Screen Shot 2019-03-11 at 11 16 33 AM

### DIFF
--- a/src/components/StaffGrid/StaffGrid.test.jsx
+++ b/src/components/StaffGrid/StaffGrid.test.jsx
@@ -1,0 +1,39 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import StaffGrid from './index';
+
+const staff = [
+  {
+    uuid: '6f23f2f8-10dd-454a-8497-2ba972c980c4',
+    given_name: 'First',
+    family_name: 'Last',
+    profile_image_url: '/media/people/profile_images/6f23f2f8-10dd-454a-8497-2ba972c980c4-a411afec9477.jpeg',
+  },
+  {
+    uuid: '17d0e2c0-9a02-421b-93bf-d081339090cc',
+    given_name: 'Pippa',
+    family_name: '',
+    profile_image_url: '/media/people/profile_images/17d0e2c0-9a02-421b-93bf-d081339090cc-68912d27b6e7.jpeg',
+  },
+  {
+    uuid: '2aba6189-ad7e-45a8-b269-bea071b80391',
+    given_name: 'Dave',
+    family_name: 'Grohl',
+    profile_image_url: '/media/people/profile_images/2aba6189-ad7e-45a8-b269-bea071b80391-11df6812f839.png',
+  },
+];
+
+describe('CourseTable', () => {
+  it('renders each member of the staff', () => {
+    const component = shallow(<StaffGrid staff={staff} />);
+    const stafferWrappers = component.find('.staffer-wrapper');
+    expect(stafferWrappers).toHaveLength(staff.length);
+
+    stafferWrappers.forEach((staffer, index) => {
+      const expected = staff[index];
+      expect(staffer.text()).toEqual(`<u /><u />${expected.given_name} ${expected.family_name}`);
+      expect(staffer.find('img').prop('src')).toEqual(expected.profile_image_url);
+    });
+  });
+});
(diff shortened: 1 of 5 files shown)


src/components/StaffGrid/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+              {staffer.position.organization_name && <div className="organization">{staffer.position.organization_name}</div>}
+            </React.Fragment>
+          }
+        </div>
+      </div>
+    ))}
+  </div>
+);
+
+StaffGrid.propTypes = {
```
bbaker6225: The prop types here are definitely excessive, I was just trying to capture all the possible data from the API, in case we do need any of this information elsewhere. Additionally, I tried to capture what seemed like required fields based off discovery admin, but I might be wrong

src/components/StaffGrid/index.jsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const StaffGrid = ({ staff }) => (
+  <div className="staff-grid row">
+    {staff.map(staffer => (
+      <div className="staffer-wrapper col-6 col-md-4 col-lg-3" key={staffer.uuid}>
+        <img src={staffer.profile_image_url} className="rounded-circle w-50" alt="" />
```
AlasdairSwan: I see you're putting the border-radius on the images themselves. While working on the updated program details pages I noticed that not all of the instructor images are in the preferred format (see the last instructor here https://sandbox-edx-prospectus.edx.org/micromasters/columbiax-business-analytics). So to prevent those ones looking broken I put the border-radius and width on the wrapper, and use the ```overflow-hidden``` class.

bbaker6225: Happy to make these changes, but on the sandbox I'm seeing the same thing as this, except with `overflow:hidden` in the css for the image wrapper

src/components/StaffGrid/index.jsx
mikix: Yeah, I usually just leave it at PropTypes.shape(), but I know there are good arguments for being more explicit.

src/components/StaffGrid/index.jsx
AlasdairSwan: Yeah, for the experiment on the sandbox I'm actually overriding the classes on ```img``` instead of changing universally. I should update that.

src/components/StaffGrid/index.jsx
```diff
@@ -13,10 +13,12 @@ const StaffGrid = ({ staff }) => (
           <Icon
             id="delete-icon"
             className={['mr-1', 'fa', 'fa-trash', 'fa-fw', 'text-danger']}
+            screenReaderText="Remove staffer"
```
AlasdairSwan: It would be more meaningful to include the staffer's name.

src/components/StaffGrid/index.jsx
```diff
@@ -13,10 +13,12 @@ const StaffGrid = ({ staff }) => (
...
           />
           <Icon
             id="edit-icon"
             className={['mr-1', 'fa', 'fa-edit', 'fa-fw']}
+            screenReaderText="Edit staffer"
```
AlasdairSwan: It would be more meaningful to include the staffer's name.

src/components/StaffGrid/index.jsx
bbaker6225: Still not completely sure what changes you want here. When I move the `rounded-circle w-50` utility classes up to the wrapper along with `overflow-hidden`, that causes some funkiness

src/components/StaffGrid/StaffGrid.test.jsx
```diff
@@ -0,0 +1,39 @@
...
+  {
+    uuid: '2aba6189-ad7e-45a8-b269-bea071b80391',
+    given_name: 'Dave',
+    family_name: 'Grohl',
+    profile_image_url: '/media/people/profile_images/2aba6189-ad7e-45a8-b269-bea071b80391-11df6812f839.png',
+  },
+];
+
+describe('CourseTable', () => {
+  it('renders each member of the staff', () => {
```
bbaker6225: This test feels more useful to me than a snapshot, but I realize I am going against the grain of the rest of publisher's tests so I'm happy to change it

bbaker6225:
@mikix @AlasdairSwan I changed the test to a snapshot to be consistent with the rest of publisher, and removed the grid from actually being used anywhere. Should be ready for a real review so @dianakhuang and I can more easily work off this for re-ordering and deleting

PR #58, opened 2019-03-08
Cleaned up tests to match expectations.

staubina:


src/data/reducers/courseInfo.test.js
```diff
@@ -18,6 +13,14 @@ describe('courseInfo reducer', () => {
     price: 100.00,
   };
 
+  beforeEach(() => {
```
Dillon-Dumesnil: Why beforeEach instead of the old way? Does one have a benefit over the other? If sticking with beforeEach, is there a way to only define initialState in the beforeEach so we don't have to have the let statement above?

src/data/reducers/courseInfo.test.js
```diff
@@ -36,7 +39,8 @@ describe('courseInfo reducer', () => {
   });
 
   it('course info receive works', () => {
-    expect(courseInfo(oldState, actions.receiveCourseInfo('test', { key: 'DemoX+TestCourse' })))
+    initialState.isFetching = true;
```
Dillon-Dumesnil: 👍 I like that. Making sure isFetching is set to false through the action. Nice

src/data/reducers/courseInfo.test.js
staubina: Before each allows me to remove a ton of duplicate code resetting that value every time.  Also reduces the general boilerplate code.

I am looking into how to just create it and set it in before each only.

staubina: As a class variable it has to be create before it is used and for scoping reasons it has to be created outside the beforeEach method.

PR #59, opened 2019-03-08
WIP Course Run Create Aj/disco 584

staubina:


src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -36,15 +33,17 @@ class CreateCourseForm extends React.Component {
       organizations,
       pristine,
       submitting,
+      isCreating,
```
jlajoie: What's the difference between submitting and isCreating?

package-lock.json
```diff
@@ -221,21 +221,21 @@
       "resolved": "https://registry.npmjs.org/@edx/paragon/-/paragon-3.9.3.tgz",
       "integrity": "sha512-jSBh8D8fCPlR3QALwQQRVXwXJUfyr0ogsI9a+PUAWBVMb4Z0nnKFIMzCWzpG8/H5N4MK3TtsgGX2XWpP0rXFzw==",
       "requires": {
-        "@edx/edx-bootstrap": "^1.0.0",
```
mikix: Why package-lock.json changes, if you didn't touch package.json?

src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -92,29 +91,49 @@ class CreateCourseForm extends React.Component {
...
+              <div>
+                Start Date
+                <span className="required">*</span>
+              </div>
+              <Field
+                name="start"
+                type="date"
+                component="input"
+                id="start-date"
+                label="Start Date"
```
mikix: Why a different label pattern here? In the rest of the Fields you pass in a React.Fragment etc. But here, you've put that bit above the Field.

Dillon-Dumesnil: You should be able to still use RenderInputTextField and pass in type="date". It's the same style as price

package-lock.json
staubina: This is not on purpose, I will look into why this happened.

src/components/CreateCoursePage/index.jsx
```diff
@@ -63,41 +74,41 @@ class CreateCoursePage extends React.Component {
...
+                        id="create-error"
+                        alertType="danger"
+                        title="Course Create Form failed to load: "
+                        message={courseInfo.error}
+                      />
+                    ) }
                   </div>
                 </div>
-              )}
+              )
```
Dillon-Dumesnil: nit: keep these on one line

src/data/actions/courseInfo.js
```diff
@@ -35,6 +38,18 @@ export function courseCreateFail(error) {
   return { type: CREATE_COURSE_FAIL, error };
 }
 
+export function createNewCourseRun(data) {
+  return { type: CREATE_COURSE_RUN, data };
+}
+
+export function courseCreateRunSuccess(data) {
```
Dillon-Dumesnil: For consistency with the constant, rename function `createCourseRunSuccess`

src/data/actions/courseInfo.js
```diff
@@ -35,6 +38,18 @@ export function courseCreateFail(error) {
...
+  return { type: CREATE_COURSE_RUN_SUCCESS, data };
+}
+
+export function courseCreateRunFail(error) {
```
Dillon-Dumesnil: For consistency with the constant, rename function `createCourseRunFail`

src/data/actions/courseInfo.js
```diff
@@ -66,15 +81,52 @@ export function fetchCourseInfo(id) {
...
+    const data = {
+      course: course.key,
+      start: courseRunData.start,
+      end: courseRunData.end,
+    };
+    return DiscoveryDataApiService.createCourseRun(data)
+      .then((response) => {
+        const courseRun = response.data;
+        dispatch(courseCreateRunSuccess(courseRun));
+        return dispatch(push(`/courses/${course.uuid}/`));
```
Dillon-Dumesnil: Shouldn't this be .../edit/?

src/data/actions/courseInfo.js
```diff
@@ -66,15 +81,52 @@ export function fetchCourseInfo(id) {
...
+      })
+      .catch((response) => {
+        let errorMsg = `Course Run create failed, please try again or contact support. \nError ( ${response.message} )`;
+        const errorData = response.response.data;
```
Dillon-Dumesnil: Should this be response.data? Just want to double check that response.response is desired.

src/data/actions/courseInfo.js
```diff
@@ -66,15 +81,52 @@ export function fetchCourseInfo(id) {
...
+        if (errorData) {
+          // More options in the error response
+          errorMsg += '\n\n';
+
+          Object.keys(errorData).forEach((errorKey) => {
+            if (Object.prototype.hasOwnProperty.call(errorData, errorKey)) {
+              errorMsg += `${errorKey}: ${errorData[errorKey]}`;
```
Dillon-Dumesnil: Maybe put a newline at the end of this for readability?

src/data/actions/courseInfo.test.js
```diff
@@ -119,3 +122,30 @@ describe('courseInfo create course actions', () => {
     expect(courseCreateFail(error)).toEqual(expectedAction);
   });
 });
+
+describe('courseInfo create course run actions', () => {
+  it('should start new course', () => {
```
Dillon-Dumesnil: Nit: test name could be more clear. Should start creating new course run

src/components/CreateCoursePage/index.jsx
```diff
@@ -17,14 +17,26 @@ class CreateCoursePage extends React.Component {
   }
 
   handleCourseCreate(options) {
+    let startDate = '';
+    let endDate = '';
+    if (Date.parse(options.start)) {
```
jlajoie: Is this necessary? If the key exists on our backend I'd assume it's a valid date. Not against it just curious about the extra defensive programming here

src/data/actions/courseInfo.js
staubina: /edit does not exist in this PR :D need to wait for yours to merge

Dillon-Dumesnil: Plan ahead! :)

src/data/actions/courseInfo.js
staubina: The new line does not work because it is not respected by the HTML block.  I think we need a better way to handle these errors in general.

src/components/CreateCoursePage/CreateCourseForm.jsx
staubina: Submitting is part of the redux-forms properties.  It switches to false before the creation actions are complete.  I spent some time trying to get it to wait for the API calls but decided to use this flag instead when I was unable to identify the issue.

src/data/actions/courseInfo.js
staubina: Going to rebase and do this once your PR merges
👍

src/components/CreateCoursePage/index.jsx
staubina: I made this field required again so it should work now without this code.  This was added while testing some error states and not including dates.

src/components/LoadingSpinner/index.jsx
```diff
@@ -1,13 +1,25 @@
...
+    </div>
     <Icon
       id="spinner"
       className={['fa', 'fa-circle-o-notch', 'fa-spin', 'fa-3x', 'fa-fw']}
     />
   </div>
 );
 
+LoadingSpinner.defaultProps = {
+  message: 'Loading...',
```
mikix: You might want the same trick used in LoadingMessage - null default and then do a `message ? message : loading...` in the html. That way someone can pass in null or empty string.

Why do we have LoadingMessage and LoadingSpinner? Feels a little overlapping.

src/data/actions/courseInfo.js
mikix: Hmm - you could make an array of error messages. Then loop over them like:

```
errorMessages.map() => {
  <div>{message}</div>
}
```

But really, we want to eventually actually tie the error with the field. Maybe add that to the bucket doc and just move on for now?

src/data/constants/courseInfo.js
```diff
@@ -7,6 +7,9 @@ const CREATE_COURSE_FAIL = 'CREATE_COURSE_FAIL';
 const EDIT_COURSE_INFO = 'EDIT_COURSE_INFO';
 const EDIT_COURSE_SUCCESS = 'EDIT_COURSE_SUCCESS';
 const EDIT_COURSE_FAIL = 'EDIT_COURSE_FAIL';
+export const CREATE_COURSE_RUN = 'CREATE_COURSE_RUN';
+export const CREATE_COURSE_RUN_SUCCESS = 'CREATE_COURSE_RUN_SUCCESS';
+export const CREATE_COURSE_RUN_FAIL = 'CREATE_COURSE_RUN_FAIL';
```
mikix: Listen. I am in favor of the inline export myself. But I believe some recommendation (some linter?) is pushing us toward the style of a large export statement at the bottom. Get on board for the big win.

Dillon-Dumesnil: I don't believe the linter cares and that it is mostly personal preference. It was referenced to me originally by Stank and reinforced by Alasdair. I also have some preference towards it since it is typically how we export all of our other files and components. Alasdair did say that he preferred it especially for files with many exports, but when there were only a few (like 3), he didn't really care one way or the other. I'm open to either, just want consistency!

src/data/actions/courseInfo.js
staubina: This is the way the data was coming from the server.  Are we doing something incorrectly when we send the response from the server side?

src/data/actions/courseInfo.js
staubina: That was my intention

staubina: I will try the div option though.

src/data/actions/courseInfo.js
mikix: Maybe? But I don't think I did anything special. Just gave DRF the exception.

src/data/constants/courseInfo.js
staubina: This was just missed after the rebase.  Changing.

PR #61, opened 2019-03-11
Add drag-and-drop functionality for Staff Grid.

dianakhuang:
I'm not sure how best to get UX review of this, but you can drag and drop instructors, which will update the `staffList` in `state` with the correct ordering.

FYI: @bbaker6225 @mikix


src/components/StaffGrid/index.jsx
```diff
@@ -10,36 +10,79 @@ const editStaff = () => {};
...
-          </button>
-          <span className="name font-weight-bold">
-            {getStafferName(staffer)}
-          </span>
-        </div>
-      </div>
-    ))}
-  </div>
-);
+export default class StaffGrid extends Component {
```
mikix: I believe eslint wants the separate line for the export.

dianakhuang:
We have tested this with keyboard movement but not with screenreaders. There is [documentation](https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/guides/screen-reader.md) that I'm reading right now that has a lot of information about how to make drag and drop screen reader friendly.

src/components/StaffGrid/index.jsx
```diff
@@ -1,7 +1,7 @@
-import React from 'react';
+import React, { Component } from 'react';
```
bbaker6225: You shouldn't need to import Component to extend it

bbaker6225:
@dianakhuang I will handle the changes around making each staff member its own component

src/components/StaffGrid/index.jsx
```diff
@@ -41,7 +41,7 @@ export default class StaffGrid extends Component {
           {provided => (
             <div ref={provided.innerRef} className="staff-grid row" {...provided.droppableProps}>
               {this.state.staffList.map((staffer, index) => (
-                <Draggable draggableId={staffer.uuid} index={index}>
+                <Draggable draggableId={staffer.given_name} index={index}>
```
bbaker6225: I don't think it's safe for us to assume that the `given_name` will be unique

dianakhuang: That's a good call. The screenreader uses the id for reading out the item it's selected, so I'm not sure `uuid` works either. I guess that's something to worry about later.

bbaker6225: Maybe use `getStafferName` for the id? Seems unlikely that they'll have the same given and family name

dianakhuang:
Per our discussion, I think we can get this reviewed + merge now and then deal with other sorts of review once this lives on a particular page and gets styled properly.

PR #63, opened 2019-03-12
Add ability to remove staff from grid

bbaker6225:


src/components/Staffer/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+const getStafferName = staffer => `${staffer.given_name} ${staffer.family_name || ''}`;
+
+const Staffer = ({ staffer, onRemove }) => (
+  <div className="staffer-wrapper col-6 col-md-4 col-lg-3 col-xl-2" key={staffer.uuid}>
+    <div className="staffer-image-wrapper overflow-hidden">
+      <img src={staffer.profile_image_url} className="rounded-circle w-50" alt="" />
+    </div>
+    <div className="staffer-details">
+      <button className="btn js-delete-btn mr-1 p-0" onClick={() => onRemove(staffer.uuid)}>
```
bbaker6225: I can see an argument for moving this out of staffer and having it on the actual grid, so I'm open to whatever

mikix: In the staffer makes sense to me.

bbaker6225:
This should be ready for real review

package.json
```diff
@@ -31,7 +31,7 @@
     "prop-types": "^15.5.10",
     "query-string": "^5.1.1",
     "react": "^16.4.2",
-    "react-beautiful-dnd": "^10.0.4",
+    "react-beautiful-dnd": "^10.1.0",
```
mikix: Was this actually necessary or just something you were in there for anyway?

dianakhuang: This was a recent update and we might as well pull it in.

dianakhuang:
FYI @mikix , I'm going to write a unit test in a separate PR.

PR #65, opened 2019-03-13
Adding dropdown to header so users can logout

Dillon-Dumesnil:


src/components/Header/index.jsx
```diff
@@ -25,11 +27,21 @@ class Header extends React.Component {
             <div className="col-md-auto">
               <Link to="/">Courses</Link>
             </div>
+            <Dropdown
+              title={this.props.username}
+              menuItems={[
+                <button onClick={() => apiClient.logout(process.env.LMS_BASE_URL)}>Logout</button>,
+              ]}
```
mikix: Ideally this would look like the Studio header, since we want to look like a part of Studio. Meaning:

1. STUDIO_BASE_URL (which does not exist yet)
2. Logout -> Sign Out
3. Add Studio Home and Maintenance links (geeze those are both weird things to have there)

I'm pro-doing-these, but I'm guessing that if you checked with AJ, he might say to not bother with parts of that, yet. So might be worth your time to check in.

src/components/Header/index.jsx
```diff
@@ -1,7 +1,9 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import { Link } from 'react-router-dom';
-import { Hyperlink } from '@edx/paragon';
+import { Dropdown, Hyperlink } from '@edx/paragon';
```
mikix: The dropdown arrow doesn't animate like it does on Studio. Is that just because paragon doesn't support that?

src/components/Header/index.jsx
```diff
@@ -25,11 +27,21 @@ class Header extends React.Component {
             <div className="col-md-auto">
               <Link to="/">Courses</Link>
             </div>
+            <Dropdown
```
mikix: Maybe add some bootstrap classes to make this float-right. It looks too weird on the left, and we know we're going to need to move it right eventually. Might as well do it now.

src/components/Header/index.jsx
Dillon-Dumesnil: Yeah, that's what it looks like. I'm actually a bit surprised it does do that in Studio.

src/components/Header/index.jsx
Dillon-Dumesnil: Settled on:
1. Adding in STUDIO_BASE_URL (pending merge in edx-internal to work in stage/prod).
2. Changed logout to Sign Out
3. Waiting for UX to determine if we should maintain those links.

mikix: Once we have STUDIO_BASE_URL, we should also go back and make the header logo go there too maybe

package-lock.json
```diff
@@ -1000,7 +1000,7 @@
     },
     "axios": {
       "version": "0.18.0",
-      "resolved": "http://registry.npmjs.org/axios/-/axios-0.18.0.tgz",
+      "resolved": "https://registry.npmjs.org/axios/-/axios-0.18.0.tgz",
```
mikix: Sure, you got it.

src/components/Header/index.jsx
```diff
@@ -16,20 +18,33 @@ class Header extends React.Component {
...
               <span className="badge badge-secondary beta">Alpha</span>
             </div>
             <div className="col-md-auto">
               <Link to="/">Courses</Link>
             </div>
+            <div className="ml-auto">
+              <Dropdown
+                title={this.props.username}
+                menuItems={[
+                  <button className="" onClick={() => apiClient.logout(process.env.LMS_BASE_URL)}>Sign Out
```
mikix: Should be STUDIO, as you mentioned IRL

PR #66, opened 2019-03-13
Add a test for drag-and-drop functionality.

dianakhuang:
This isn't the most beautiful thing in the world, and it isn't as black-box as I'd like it to be, but it verifies that we can move things around in the list.


dianakhuang:
I feel like there's some more mysterious magic that I needed to mock out or figure out how to manipulate better. It just didn't seem worth constantly running against a wall to get it over the line, so I decided that this was going to be the best alternative.

PR #68, opened 2019-03-15
Only show mode and price field if an entitlement exists in the course…

Dillon-Dumesnil:
…. Still need to fix discovery because you still can't switch modes


src/components/EditCoursePage/index.jsx
```diff
@@ -55,7 +55,7 @@ class EditCoursePage extends React.Component {
       courseData.subjectSecondary,
       courseData.subjectTertiary,
     ].filter(subject => !!subject);
-    courseData.entitlements = entitlements && [{
+    courseData.entitlements = entitlements[0] && [{
```
jlajoie: You'll get an error for this if entitlements is undefined

PR #73, opened 2019-03-22
WIP: Add an autocomplete action to the service.

dianakhuang:
FYI @mikedikan


src/data/services/DiscoveryDataApiService.js
```diff
@@ -2,6 +2,7 @@ import apiClient from '../apiClient';
 
 class DiscoveryDataApiService {
   static discoveryBaseUrl = `${process.env.DISCOVERY_API_BASE_URL}/api/v1`;
+  static autocompleteBaseUrl = `${process.env.DISCOVERY_API_BASE_URL}/admin/course_metadata`;
```
mikedikan: I'm guessing this URL is st up this way because this is not a DRF endpoint?  Do any other endpoints exist under this base url?

src/data/services/DiscoveryDataApiService.js
```diff
@@ -56,6 +57,17 @@ class DiscoveryDataApiService {
     // PATCH to Course endpoint to create
     return apiClient.patch(url, courseData);
   }
+
+  static autocompletePerson(text, orgs) {
```
mikedikan: do any of the other actions have tests?

src/data/services/DiscoveryDataApiService.js
dianakhuang: Yes, it's not one of the ones that uses `/api/v1`. The other endpoints that also use this base url are mostly other autocomplete endpoints like `course-autocomplete` and `organization-autocomplete`.

PR #75, opened 2019-03-26
DISCO-546: Adds in Course Run Edit Form

jlajoie:
Still need to adds in tests, also have some cherry picked code from the Instructors PR


src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
```diff
@@ -0,0 +1,208 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Field, FieldArray } from 'redux-form';
+import moment from 'moment';
+
+import { Collapsible } from '@edx/paragon';
```
bbaker6225: Super nit: This should go with the top group of imports

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
```diff
@@ -0,0 +1,208 @@
...
+import RenderInputTextField from '../RenderInputTextField';
+import RenderSelectField from '../RenderSelectField';
+import TranscriptLanguage from './TranscriptLanguage';
+
+const formatCourseRunTitle = (courseRun) => {
+  if (courseRun) {
+    const startDateString = moment(courseRun.start).format('MMM Do YYYY');
+    const pacingTypeString = courseRun.pacing_type.split('_').map(pacingType =>
```
bbaker6225: This line is a chore to parse. But it's fine™️

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
```diff
@@ -0,0 +1,208 @@
...
+  handleRemove(index) {
+    this.props.fields.remove(index);
+  }
+
+  render() {
+    const { fields, courseRuns, languageOptions } = this.props;
+    return (
+      <div>
+        {fields.map((courseRun, index) => (
+          <Collapsible
```
bbaker6225: For a11y you should add an `iconId` attribute to the collapsible that is just something like `collapsible-icon-${index}` to avoid an issue with paragon's dumb id duplication

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
```diff
@@ -0,0 +1,208 @@
...
+          >
+            <Field
+              name={`${courseRun}.start`}
+              type="date"
+              component={RenderInputTextField}
+              format={value => moment(value).format('YYYY-MM-DD')}
+              label={
+                <React.Fragment>
+                  Start date
+                  <span className="required">*</span>
```
bbaker6225: Copying Alasdair's comment on my PR:  `a11y: the * should reference text at the bottom of the form explaining that this indicates that it is required.`

For all of the required fields with the asterisk

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
```diff
@@ -0,0 +1,208 @@
...
+                </React.Fragment>
+              }
+              required
+            />
+            <Field
+              name={`${courseRun}.pacing_type`}
+              type="text"
+              component={RenderSelectField}
+              options={
+                [
```
bbaker6225: ultra nit: having the options inline looks super ugly to me, but whatever not worth changing

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
```diff
@@ -0,0 +1,208 @@
...
+                  Maximum Effort
+                </React.Fragment>
+              }
+            />
+            <Field
+              name={`${courseRun}.pacing_type`}
+              type="text"
+              component={RenderSelectField}
+              options={
+                [
```
bbaker6225: Never mind, does seem worthwhile to move this to a variable or function since it's used in multiple places

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -45,18 +63,22 @@ class EditCourseForm extends React.Component {
...
+    const languageOptions = courseRunOptions &&
+    this.parseOptions(courseRunOptions.content_language.choices);
 
     levelTypeOptions.unshift({ label: '--', value: '' });
     subjectOptions.unshift({ label: '--', value: '' });
 
     return (
       <div className="edit-course-form row">
-        <div className="col">
+        <div className="row col">
```
bbaker6225: It's sort of weird to have a `row` and `col` on the same class

src/components/EditCoursePage/TranscriptLanguage.jsx
```diff
@@ -0,0 +1,64 @@
...
+                onRemove={this.handleRemove}
+                targetFieldNumber={index}
+              />
+            </li>
+          ))}
+        </ul>
+        <button
+          type="button"
+          className="btn btn-light js-add-button"
+          onClick={() => fields.push('')}
```
bbaker6225: I think you need to push `{}`? But I might be wrong

src/components/EditCoursePage/index.jsx
```diff
@@ -45,11 +46,29 @@ class EditCoursePage extends React.Component {
...
       editCourse,
     } = this.props;
     const courseData = options;
+    const modifiedCourseRuns = [];
+    const newCourseRuns = [];
+    courseData.course_runs.forEach((modifiedCourseRun) => {
+      let found = false;
+      initialCourseRuns.forEach((initialCourseRun) => {
+        if (initialCourseRun.key === modifiedCourseRun.key) {
+          found = true;
```
bbaker6225: The `found` makes me think that `.some()` would be useful here, but maybe it doesn't work cleanly with pusing multiple runs

src/components/EditCoursePage/index.jsx
```diff
@@ -116,6 +137,7 @@ class EditCoursePage extends React.Component {
     const entitlement = entitlements && entitlements[0];
     const mode = entitlement && entitlement.mode;
     const price = entitlement && entitlement.price;
+    const courseRuns = (course_runs && course_runs.length > 0) ? course_runs : [];
```
bbaker6225: You shouldn't need the check that length is greater than 0 if you're defaulting to `[]` right?

package.json
```diff
@@ -28,6 +28,7 @@
     "email-prop-type": "^3.0.0",
     "font-awesome": "^4.7.0",
     "history": "^4.7.2",
+    "moment": "^2.24.0",
```
mikix: I checked out this npm page - 32k dependents on moment. That's a widely used library...

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
jlajoie: Will add a comment, but it seemed better than doing a switch(pacing_type): case 'self_paced':

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
jlajoie: Ooooh nice! Thanks! And we're cool with the index being included in the unique id? I think our linter typically yells at us for that

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
jlajoie: 👍

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
bbaker6225: Yeah, you shouldn't use index for the `key`, but it's fine for this

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
bbaker6225: The resolution from Jeff Witt was just to put `aria-hidden` on the spans with the asterisks and make sure we're passing the `required` attribute, which you're doing

src/components/EditCoursePage/index.jsx
mikix: Do you need the length check?

src/components/EditCoursePage/index.jsx
```diff
@@ -45,11 +46,29 @@ class EditCoursePage extends React.Component {
...
     } = this.props;
     const courseData = options;
+    const modifiedCourseRuns = [];
+    const newCourseRuns = [];
+    courseData.course_runs.forEach((modifiedCourseRun) => {
+      let found = false;
+      initialCourseRuns.forEach((initialCourseRun) => {
+        if (initialCourseRun.key === modifiedCourseRun.key) {
+          found = true;
+          if (JSON.stringify(initialCourseRun) !== JSON.stringify(modifiedCourseRun)) {
```
mikix: React forms has some meta thing that might be able to tell you if it was modified. I think meta.touched? @bbaker6225 might know more.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -34,6 +36,22 @@ class EditCourseForm extends React.Component {
...
+    }
+
+    const { data } = courseRunOptions;
+
+    if (!data || !data.actions) {
+      return [];
+    }
+
+    return data.actions.POST;
+  }
```
mikix: Blegh. It's appropriate, but blegh.

src/components/EditCoursePage/index.jsx
bbaker6225: You should be able to receive the prop `meta: { touched, error }`  on individual fields, but I don't think it will exist at the page level

mikix: But maybe the form could expose that? Or you could just iterate on the form and see. Stringify works too, just feels less Form-y.

PR #80, opened 2019-03-28
Add staff list to edit course

bbaker6225:
My initial approach with field arrays was wrong and silly, so I undid everything. The staff list inside the collapsible doesn't look awesome, but it's ok


src/components/EditCoursePage/__snapshots__/CollapsibleCourseRunFields.test.jsx.snap
```diff
@@ -68,7 +68,7 @@ ShallowWrapper {
                 className="course-run-label"
               >
                 <span>
-                  Jan 1st 2000 - Self Paced
+                  Dec 31st 1999 - Self Paced
```
bbaker6225: My snapshots don't like Y2k for some reason, but I don't know why

bbaker6225:
:works_on_my_machine:
I'll take a look Monday and see what's going on with yours

src/components/EditCoursePage/CollapsibleCourseRunFields.test.jsx
```diff
@@ -19,7 +19,7 @@ const pacingTypeOptions = [
 
 const initialvalues = [
   {
-    start: '2000-01-01T00:00:00Z', // Different format to be transformed
+    start: '2000-01-01T12:00:00Z', // Different format to be transformed
```
bbaker6225: Change it to 12 so that you don't get different days depending on if you run the test inside or outside the shell

src/components/EditCoursePage/index.jsx
```diff
@@ -55,24 +64,19 @@ class EditCoursePage extends React.Component {
     const modifiedCourseRuns = [];
     const newCourseRuns = [];
     courseData.course_runs.forEach((modifiedCourseRun) => {
-      let found = initialCourseRuns.some(initialCourseRun => (
+      const found = initialCourseRuns.some(initialCourseRun => (
         initialCourseRun.key === modifiedCourseRun.key &&
         JSON.stringify(initialCourseRun) !== JSON.stringify(modifiedCourseRun)
       ));
-      initialCourseRuns.forEach((initialCourseRun) => {
```
bbaker6225: Duplicated logic that was accidentally left in during the edit course runs PR

bbaker6225:
You (probably) won't be able to run this as is locally due to some issues with the edit course page and the `course_run` api that Jeff and AJ are looking into, but if you want to copy my hacks that let you test out the staff part, let me know

PR #84, opened 2019-04-02
DISCO-777: Reduces PATCH body for course runs to only form fields

jlajoie:


src/data/services/DiscoveryDataApiService.js
```diff
@@ -73,7 +73,18 @@ class DiscoveryDataApiService {
     // Create a promises array to handle all of the new/modified course runs
     const promises = courseRunData.map((courseRun) => {
       const courseRunUrl = `${DiscoveryDataApiService.discoveryBaseUrl}/course_runs/${courseRun.key}/`;
-      return apiClient.patch(courseRunUrl, courseRun);
+      // Minimal object needed to be written to Course Runs endpoint
+      const courseRunPatchBody = {
```
bbaker6225: I feel like there has to be a better way to send only the fields changed to discovery

src/components/EditCoursePage/index.jsx
```diff
@@ -69,13 +69,16 @@ class EditCoursePage extends React.Component {
       modifiedCourseRun.start = moment(modifiedCourseRun.start).toISOString();
       modifiedCourseRun.end = moment(modifiedCourseRun.end).toISOString();
       /* eslint-enable no-param-reassign */
-      const found = initialCourseRuns.some(initialCourseRun => (
```
bbaker6225: Still not seeing why this doesn't work, but I trust ya

bbaker6225:
"Check which fields are dirty and only send those" - You

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
```diff
@@ -47,136 +47,119 @@ const formatCourseRunTitle = (courseRun) => {
...
-          type="button"
-          className="btn btn-light js-add-button"
-          onClick={() => fields.push({})}
+const CollapsibleCourseRunFields = (props) => {
+  const {
+    fields,
+    courseRuns,
+    languageOptions,
+    pacingTypeOptions,
+  } = props;
```
bbaker6225: Nit: Seems a little cleaner to destructure the props right when they're passed in like https://github.com/edx/publisher-frontend/blob/d57ff232fd11713e8ef35e1e3093d8b0f92c017c/src/components/RemoveButton/index.jsx#L5-L10

src/components/EditCoursePage/index.jsx
```diff
@@ -2,12 +2,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Helmet from 'react-helmet';
-import moment from 'moment';
 
 import EditCourseForm from './EditCourseForm';
 import StatusAlert from '../StatusAlert';
 import LoadingSpinner from '../LoadingSpinner';
-import { getCourseNumber, jsonDeepCopy } from '../../utils';
```
bbaker6225: I think we still need the deep copy on the course runs since https://github.com/edx/publisher-frontend/blob/d57ff232fd11713e8ef35e1e3093d8b0f92c017c/src/components/EditCoursePage/index.jsx#L36-L40 directly modifies the staff

src/components/EditCoursePage/index.jsx
```diff
@@ -58,31 +57,13 @@ class EditCoursePage extends React.Component {
...
-        initialCourseRun.key === modifiedCourseRun.key &&
-        JSON.stringify(initialCourseRun) !== JSON.stringify(modifiedCourseRun)
-      ));
-      if (found) {
-        modifiedCourseRuns.push(jsonDeepCopy(modifiedCourseRun));
-      } else {
-        newCourseRuns.push(jsonDeepCopy(modifiedCourseRun));
-      }
-    });
+    const courseData = values;
```
bbaker6225: Nit: why not just call `values` `courseData` in the first place?

src/components/EditCoursePage/index.jsx
```diff
@@ -138,6 +119,17 @@ class EditCoursePage extends React.Component {
       courseRunOptions,
     } = this.props;
     const { startedFetching } = this.state;
+    const minimalCourseRuns = course_runs && course_runs.map(courseRun => ({
```
bbaker6225: I think you need staff in the fields here? Hopefully that will fix the issue I'm seeing where staff aren't loaded even when they're returned from discovery

jlajoie: 👍

src/components/EditCoursePage/index.jsx
jlajoie: Ahh gotcha, I actually messed up and rebased out the staff work originally so thanks for the catch here

bbaker6225:
Actually ran it, still looks good

PR #85, opened 2019-04-02
Fix issue with staff on new course run

bbaker6225:


src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
```diff
@@ -159,12 +160,18 @@ class CollapsibleCourseRunFields extends React.Component {
...
+            />
           </Collapsible>
         ))}
         <button
           type="button"
           className="btn btn-light js-add-button"
-          onClick={() => fields.push({})}
+          onClick={() => fields.push({
+            staff: [],
+          })}
```
mikix: I don't have any context for this change - so I don't know why this would fix anything. But sure?

bbaker6225: When you add a new course run, staff gets initialized to an empty string by default, so this is sort of a hack to make sure it's always in array form

mikix: All fields get initialized to empty string? Do we have the wrong default or field type somewhere? Or are fields supposed to interpret empty string as null value and set themselves internally correctly?

bbaker6225: All fields get passed value as a string for initialization. From the docs I'm seeing this example where the deal with value being an array, so I'll try and handle it like they do
```javascript
const renderMultiselect = ({ input, data, valueField, textField }) =>
  <Multiselect {...input}
    onBlur={() => input.onBlur()}
    value={input.value || []} // requires value to be an array
    data={data}
    valueField={valueField}
    textField={textField}
  />
```

PR #86, opened 2019-04-03
Adding Course Run Create form and button

staubina:
Added the Create Course run form. Planning to rebase on the upcoming changes.

### DIFF
--- a/src/components/CreateCourseRunPage/CreateCourseRunForm.jsx
+++ b/src/components/CreateCourseRunPage/CreateCourseRunForm.jsx
@@ -1,0 +1,96 @@
+import React from 'react';
+import { Field, reduxForm } from 'redux-form';
+import PropTypes from 'prop-types';
+import { Link } from 'react-router-dom';
+
+import RenderInputTextField from '../RenderInputTextField';
+
+const BaseCreateCourseRunForm = ({
+  handleSubmit,
+  pristine,
+  submitting,
+  title,
+  uuid,
+}) => {
+  const formControlDisabled = pristine || submitting;
+  return (
+    <React.Fragment>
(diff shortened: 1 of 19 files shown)


src/components/CreateCourseRunPage/CreateCourseRunForm.jsx
```diff
@@ -0,0 +1,98 @@
...
+  );
+};
+
+BaseCreateCourseRunForm.defaultProps = {
+};
+
+BaseCreateCourseRunForm.propTypes = {
+  handleSubmit: PropTypes.func.isRequired,
+  initialValues: PropTypes.shape({ // eslint-disable-line react/no-unused-prop-types
+  }).isRequired,
```
mikix: You coud do `shape().isRequired` I think - might avoid that eslint warning?

staubina:
As for Snapshot tests, I did take some time looking at the snapshots and making sure what I expected was there.

The interaction tests seemed to not be of much value due to the lack of POST test options.  We still need to dig into that and how to get POST tests in place.

src/components/CreateCourseRunPage/CreateCourseRunForm.jsx
staubina: I will give this a try.

staubina: I actually ended up removing it.  I realized I Was not using it @mikix

mikix: Heh, also a good solution. :)

staubina:
Tweaked ui a bit from your comments @mikix 

Rebasing for a merge

PR #89, opened 2019-04-05
updated Create Course Page and tests

staubina:
Updates to the Create Course Page to new format to better support snapshot tests.

### DIFF
--- a/src/components/CreateCoursePage/CreateCourseForm.jsx
+++ b/src/components/CreateCoursePage/CreateCourseForm.jsx
@@ -6,7 +6,7 @@ import RenderInputTextField from '../RenderInputTextField';
 import RenderSelectField from '../RenderSelectField';
 import { AUDIT_TRACK, VERIFIED_TRACK, PROFESSIONAL_TRACK } from '../../data/constants';
 
-class CreateCourseForm extends React.Component {
+class BaseCreateCourseForm extends React.Component {
   getEnrollmentTrackOptions() {
     return [
       { label: 'Select enrollment track', value: '' },
@@ -131,7 +131,7 @@ class CreateCourseForm extends React.Component {
               <div className="row justify-content-end">
                 <button
                   type="submit"
-                  className="btn btn-outline-primary form-submit-btn"
+                  className="btn btn-primary form-submit-btn"
                   disabled={pristine || submitting || isCreating}
                 >
                   Submit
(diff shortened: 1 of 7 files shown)


src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -169,7 +169,8 @@ CreateCourseForm.propTypes = {
   isCreating: PropTypes.bool,
 };
 
-export default reduxForm({
+const CreateCourseForm = reduxForm({
```
bbaker6225: I would do `export default CreateCourseForm` and then do a regular export on the base, that way you only have to update imports on the test file

src/components/CreateCoursePage/CreateCourseForm.test.jsx
```diff
@@ -1,18 +1,28 @@
...
+const organizations = [
+  { name: 'edX', key: 'edx' }, { name: 'edX2', key: 'edx2' },
+];
 
 describe('CreateCourseForm', () => {
   it('renders html correctly with no orgs', () => {
-    const component = shallow(<CreateCourseForm />);
+    const component = shallow(<BaseCreateCourseForm
+      handleSubmit={() => {}}
+      pristine
```
bbaker6225: pristine and submitting seem like they should be default props like I did here, but personal opinion https://github.com/edx/publisher-frontend/blob/5c4253d9f6e2450f0de3425b5015ef0067055d4e/src/components/EditCoursePage/EditCourseForm.jsx#L268-#L272

src/components/CreateCoursePage/CreateCourseForm.jsx
staubina: 👍 Was following the other examples.  I can adjust this.

src/components/CreateCoursePage/CreateCourseForm.test.jsx
staubina: I will take a look.

src/components/CreateCoursePage/CreateCourseForm.jsx
bbaker6225: Yeah, that's how @inventhouse and I did it on the `CreateInstructorForm`, but we should have done export default on the connected one

bbaker6225: Sorry for the confusion

staubina: No worries.

src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -164,13 +166,14 @@ BaseCreateCourseForm.propTypes = {
...
+  pristine: PropTypes.bool,
+  submitting: PropTypes.bool,
   isCreating: PropTypes.bool,
 };
 
 const CreateCourseForm = reduxForm({
   form: 'create-course-form',
 })(BaseCreateCourseForm);
-export { BaseCreateCourseForm, CreateCourseForm };
+export default CreateCourseForm;
```
bbaker6225: no need to change just an fyi, you can just do export default CreateCourseForm = reduxForm...

staubina: Ok, I thought it would be more clear to have both exports below.  I will adjust.

staubina: Need to rebase anyway

PR #97, opened 2019-04-09
Disable fields when a course run is in review.

dianakhuang:
https://openedx.atlassian.net/browse/DISCO-757

Disables all course fields if any of the course runs are in review.

### DIFF
--- a/src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
+++ b/src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
@@ -137,6 +137,7 @@ const CollapsibleCourseRunFields = ({
           <button
             type="submit"
             className="btn btn-primary form-submit-btn"
+            disabled={courseInReview}
           >
             {courseRun.status === 'published' ? (
               <span>Publish</span>
--- a/src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
+++ b/src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
@@ -54,6 +54,7 @@ const CollapsibleCourseRunFields = ({
   courseRuns,
   languageOptions,
   pacingTypeOptions,
+  courseInReview,
 }) => (
   <div>
     {fields.map((courseRun, index) => (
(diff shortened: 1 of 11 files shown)


src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
```diff
@@ -74,6 +75,7 @@ const CollapsibleCourseRunFields = ({
           }
           placeholder="mm/dd/yyyy"
```
mikix: Is that really our placeholder? That's not very non-US friendly. I'd assumed when I saw that on the page, that it was coming from the browser.

Obviously not your bug or problem, but humph.

dianakhuang:
Whoops, course runs got lost in conflict resolution. Thanks for catching that!

I was thinking about getting the background gray, but considering how these lines are spread all over the place, I wanted to get this part in, at least, so that I'm not constantly trying to go through conflict resolution (and break all the things).

mikix:
Speaking of: sorry but I think I just caused you some conflicts with my field label changes

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
dianakhuang: I think if we're pulling in `moment` anyway, we should be able to ask `moment` for a reasonable placeholder.

mikix:
The StaffList component isn't being disabled properly (presumably needs some custom disabling logic?)

mikix:
Disabling the staff list was difficult? Also, the transcript language widgets are still enabled. Plus all the buttons on the page should be disabled, yeah?

dianakhuang:
The staff list is a bit tricky, so I'm going to handle that separately, since we'll have to toggle the remove /draggable/add/edit buttons, too.  Same with the language widget. The buttons were not specified, but maybe I should have assumed that.

dianakhuang:
It's good to be calling these things out! I really do appreciate the thoroughness, especially when I forget to put things on my checklist.

PR #99, opened 2019-04-10
Add FieldLabel class and use it on every field

mikix:
This standardizes the look of field labels. I also corrected some spacing where I could between fields.

I'm still ignoring the instructor page, since it's about to be refactored.

### DIFF
--- a/src/components/FieldLabel/index.jsx
+++ b/src/components/FieldLabel/index.jsx
@@ -1,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const FieldLabel = (props) => {
+  let requireText = '';
+  if (props.required) {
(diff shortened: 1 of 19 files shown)


src/components/FieldLabel/index.jsx
```diff
@@ -0,0 +1,33 @@
...
+      <strong>{props.text}:</strong>
+      {requireText}
+    </div>
+  );
+};
+
+FieldLabel.defaultProps = {
+  className: '',
+  required: false,
+  requiredForSubmit: false,
```
bbaker6225: You don't have `requiredForSave` in the `propTypes`, but you do use it somewhere else. Maybe a mixup in language?

mikix: Yes, thank you. I renamed a property last minute. See! Reviews are good after all.

PR #102, opened 2019-04-11
Updates the page size request to get more organizations on create

staubina:


src/data/services/DiscoveryDataApiService.js
```diff
@@ -37,8 +37,15 @@ class DiscoveryDataApiService {
...
 
   static fetchOrganizations() {
+    const queryParams = {
+      page: 1,
+      page_size: 500,
+      limit: 500,
+    };
     const url = `${DiscoveryDataApiService.discoveryBaseUrl}/organizations/`;
-    return apiClient.get(url);
+    return apiClient.get(`${url}`, {
```
jlajoie: Don't interpolate just a string here 😄

src/data/services/DiscoveryDataApiService.js
```diff
@@ -37,8 +37,15 @@ class DiscoveryDataApiService {
   }
 
   static fetchOrganizations() {
+    const queryParams = {
+      page: 1,
+      page_size: 500,
+      limit: 500,
+    };
```
mikix: This assumes we only have 500 orgs... 

When is this used? Can we add a flag that's like - show me only my orgs?

PR #103, opened 2019-04-11
update styling for x

staubina:


src/components/RemoveButton/index.jsx
```diff
@@ -17,7 +17,7 @@ const RemoveButton = ({
     >
       <Icon
         id="remove-field"
-        className={['fa', 'fa-close', 'fa-2x', 'text-danger']}
+        className={['fa', 'fa-close']}
```
bbaker6225:
```suggestion
        className={['fa', 'fa-close', 'fa-5x', 'text-danger', 'bg-success']}
```


Screen Shot 2019-04-11 at 9 42 07 AM
I think this is what we want

PR #105, opened 2019-04-11
Style rich text editor components to look like other disabled fields.

dianakhuang:
https://openedx.atlassian.net/browse/DISCO-757

This is mostly a visual cleanup to make the fields look more consistent.


dianakhuang:
I tagged you @AlasdairSwan because I would appreciate knowing if this is too hacky or if there's a better way to do what I'm trying to do.

PR #106, opened 2019-04-11
Only show error message if there is an error.

Dillon-Dumesnil:


src/components/EditCoursePage/index.jsx
```diff
@@ -85,7 +85,7 @@ class EditCoursePage extends React.Component {
   }
 
   render() {
-    if (!this.props.courseInfo || !this.props.courseOptions) {
+    if (!this.props.courseInfo || !this.props.courseOptions || !this.props.courseRunOptions) {
       return (
         <StatusAlert
           id="error"
```
staubina: We should add an action the user can take.

For example, `Please try to reload the page and if this does not work contact support`.

src/components/EditCoursePage/index.jsx
```diff
@@ -157,11 +157,15 @@ class EditCoursePage extends React.Component {
...
+      error = error.concat(courseRunOptions.error);
     }
     error = error.trim();
     const showSpinner = !startedFetching || courseInfo.isFetching || courseOptions.isFetching;
-    const showForm = startedFetching && !courseInfo.isFetching && !courseOptions.isFetching;
+    const showForm = !error && startedFetching && !courseInfo.isFetching &&
+      !courseOptions.isFetching;
 
     return (
       <React.Fragment>
```
staubina: Same comment as above, should we provide next steps to the users in the error message below?

src/components/EditCoursePage/index.jsx
Dillon-Dumesnil: Happy to add that in, but if this scenario happens, I think we are in a bigger mess. This error messages when the entire object does not exist i.e. none of the `{data: {}, isFetching: False, error: null}`

staubina: The main thing is contact support.  In the future, we should add a direct link to support of some fashion.  Even just an email.

Dillon-Dumesnil:
@staubina Updated the error messaging to include a message about retrying and contacting support if it persists. This will show up in either error case.

PR #107, opened 2019-04-11
Upgrade frontend-auth to 4.0.0

douglashall:
The 4.0.0 version of frontend-auth resolves an issue where users would be prompted to login more frequently than necessary when reloading the app.


staubina:
Also, did something change with how the API Client works? Looks like our tests are now failing.

PR #116, opened 2019-04-16
Making autosuggest item list expand container dynamically.

mikedikan:
DISCO-824

### DIFF
--- a/src/components/EditCoursePage/CollapsibleCourseRunFields.scss
+++ b/src/components/EditCoursePage/CollapsibleCourseRunFields.scss
@@ -1,0 +1,3 @@
+.collapsible-body {
+  overflow: visible !important;
+}
--- a/src/components/StaffList/StaffList.scss
+++ b/src/components/StaffList/StaffList.scss
@@ -1,6 +1,5 @@
 .react-autosuggest__container {
   position: relative;
-  margin-bottom: 300px;
 }
 
 .react-autosuggest__input {
@@ -32,9 +31,10 @@
   position: absolute;
   top: 29px;
   width: 100%;
-  border: 1px solid $medium-gray;
-  margin-top: 0;
+  border: 1px solid $light-gray;
+  margin-top: 12px;
   z-index: 2;
+  background-color: $white;
   border-bottom: {
     left-radius: 0.1875rem;
     right-radius: 0.1875rem;
@@ -45,11 +45,10 @@
   margin: 0;
   padding: 0;
   list-style-type: none;
-  max-height: 300px;
+  max-height: 318px;
   overflow-y: auto;
 }
 
 .react-autosuggest__suggestion--highlighted {
   background-color: $light-gray;
 }
-
--- a/src/components/StaffList/index.jsx
+++ b/src/components/StaffList/index.jsx
@@ -201,7 +201,6 @@ class StaffList extends React.Component {
 
     return (
       <React.Fragment>
-        Staff
         <DragDropContext onDragEnd={this.onDragEnd}>
           <Droppable droppableId="StaffList" direction="vertical">
             {provided => (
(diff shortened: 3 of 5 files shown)


src/components/StaffList/index.jsx
```diff
@@ -233,7 +235,8 @@ class StaffList extends React.Component {
           </Droppable>
         </DragDropContext>
         {/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */}
-        <label className="col-12" id="label-staff-search" htmlFor="staff-search" onKeyDown={this.handleAutosuggestEnterEvent}>Search or add new instructor
+        <label className={suggestionVisible ? 'w-100 suggestions-visible' : 'w-100'} id="label-staff-search" htmlFor="staff-search" onKeyDown={this.handleAutosuggestEnterEvent}>
```
dianakhuang: There's a library called `classNames` that is already used in publisher-frontend that is a cleaner way to handle this ternary operator.

mikedikan:
I created https://openedx.atlassian.net/browse/DISCO-827 to deal with pop-up or pop-down question.

src/components/StaffList/index.jsx
mikedikan: Don't need to write the class dynamically anymore, but good to know!

src/sass/App.scss
```diff
@@ -9,6 +9,7 @@ $fa-font-path: "~font-awesome/fonts";
 @import "../components/RichEditor/RichEditor.scss";
 // Required to have the Collapsible component work
 @import '../../node_modules/@edx/paragon/src/Collapsible/Collapsible';
+@import '../components/EditCoursePage/CollapsibleCourseRunFields';
```
bbaker6225: Nit: our scss files should be grouped together apart from the `node_modules` ones

mikedikan: cleaned the file up a bit

src/sass/App.scss
```diff
@@ -1,15 +1,17 @@
...
 @import "./_variables.scss";
+
 @import "../../node_modules/@edx/edx-bootstrap/sass/edx/theme";
 @import "../../node_modules/bootstrap/scss/bootstrap";
 @import "../../node_modules/@edx/paragon/src/index";
 @import "../../node_modules/font-awesome/scss/font-awesome";
 @import "../../node_modules/@edx/frontend-component-footer/src/lib/scss/site-footer";
-@import "../components/RichEditor/RichEditor.scss";
 // Required to have the Collapsible component work
 @import '../../node_modules/@edx/paragon/src/Collapsible/Collapsible';
```
mikix: The two lines above were removed by me in a prevoius PR - looks like you added it back in one of your merges? Can you drop them here again, while you're there?

Also, all of our imports are now relative with the `../../node_modules`? They should just be doing `~` or relative to `node_modules` and our webpack config should look in `node_modules` for them... Looks like a change you made in a previous PR? I guess you don't need to clean that bit up here, but it's not our preferred import strategy.

src/components/EditCoursePage/CollapsibleCourseRunFields.scss
```diff
@@ -0,0 +1,3 @@
+.collapsible-body {
+  overflow: visible !important;
+}
```
mikix: I was looking for whether we could just do this in a bootrstrap class, and didn't see anything in my quick scan. But I did came across:

https://getbootstrap.com/docs/4.0/components/dropdowns/#options

Might be useful when we get to the "dropping up" part of this work.

src/components/StaffList/StaffList.scss
```diff
@@ -45,11 +45,10 @@
...
   padding: 0;
   list-style-type: none;
-  max-height: 300px;
+  max-height: 318px;
   overflow-y: auto;
 }
 
 .react-autosuggest__suggestion--highlighted {
   background-color: $light-gray;
 }
```
mikix: Ugh, we have so much custom scss here. I suspect a lot of it could be replaced with bootstrap - like the width, the absolute position, the border stuff. Doesn't have to be cleaned up in this PR. But just a note that custom scss is more difficult to maintain. Please consider avoiding it when possible.

mikedikan: spoke in person about custom css and 3rd party components

PR #118, opened 2019-04-17
Fixing the issue where saving would result in a white page being disp…

Dillon-Dumesnil:
…layed. This was caused because instead of only returning a single promise with the course data, we return an array with the course run data followed by the course data.


src/data/actions/courseInfo.js
```diff
@@ -156,7 +156,7 @@ function editCourse(courseData, courseRunData) {
     // Send edit course PATCH
     return DiscoveryDataApiService.editCourse(courseData, courseRunData)
       .then((response) => {
-        const course = response.data;
+        const course = response[1].data;
```
jlajoie: You'll want to grab the last element for the array 🙃

PR #121, opened 2019-04-18
Adding status pills to course and course run components

Dillon-Dumesnil:


Dillon-Dumesnil:
image

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
```diff
@@ -5,13 +5,15 @@ import { Field, FieldArray, submit } from 'redux-form';
 import { Collapsible } from '@edx/paragon';
 import { connect } from 'react-redux';
 
+import Pill from '../Pill';
```
mikix: Unrelated to your perfect change here - but the fact that our style linter doesn't enforce alphabetical imports is driving me batty. Look at this list! It's nonsense.

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
```diff
@@ -25,6 +27,11 @@ const formatCourseRunTitle = (courseRun) => {
     return (
       <div className="course-run-label">
         <span>{`Course run starting on ${labelItems.join(' - ')}`}</span>
+        {/*
+          After we have a way of determining if the course run has been edited, that should be
+          added into the list of statuses being passed into the Pill component.
+        */}
```
mikix: nit: I'd toss a `TODO:` on the front of that - makes it a pop a little more - scanning this comment would make me think it's just normal dev comments about what the code is doing.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -62,6 +63,17 @@ export class BaseEditCourseForm extends React.Component {
       ({ label: choice.display_name, value: choice.value }));
   }
 
+  formatCourseTitle(title, courseStatuses) {
+    // After we have a way of determining if the course has been edited, that should be
+    // added into the list of statuses being passed into the Pill component.
```
mikix: Ditto.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -93,10 +106,10 @@ export class BaseEditCourseForm extends React.Component {
     return (
       <div className="edit-course-form">
         <form onSubmit={handleSubmit} id={id}>
+          <FieldLabel text="Course" className="mt-4 mb-2" />
           <Collapsible
-            title={`Course: ${title}`}
+            title={this.formatCourseTitle(title, courseStatuses)}
             key="Test Key"
-            isOpen
```
mikix: Oh yeah?

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -93,10 +106,10 @@ export class BaseEditCourseForm extends React.Component {
     return (
       <div className="edit-course-form">
         <form onSubmit={handleSubmit} id={id}>
+          <FieldLabel text="Course" className="mt-4 mb-2" />
```
mikix: Is this Dillon UI or Carol UI? It's good either way, just curious.

src/components/StafferPage/index.jsx
```diff
@@ -39,6 +39,7 @@ class StafferPage extends React.Component {
       ({
         ...link,
         id: link.id || null,
+        title: link.title || '',
```
mikix: :+1:

src/components/Pill/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+    </span>
+  ));
+};
+
+const Pill = ({
+  statuses,
+}) => getPills(statuses);
+
+Pill.propTypes = {
+  statuses: PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.bool, PropTypes.string])).isRequired,
```
mikix: Why allow bool here? Doesn't look like that case is handled in `render`?

src/components/Pill/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+  return pills.map(pill => (
+    <span className={classNames('ml-2', pill.className)}>
+      {pill.text}
+    </span>
+  ));
+};
+
+const Pill = ({
+  statuses,
+}) => getPills(statuses);
```
mikix: Why have both `Pill` and `getPills`? They seem the same.

src/components/Pill/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+        break;
+      case 'review_by_legal':
+      case 'review_by_internal':
+        pills.push({
+          text: 'In review',
+          className: 'badge badge-warning',
+        });
+        pills.push({
+          text: <i className="fa fa-lock" />,
+        });
```
mikix: Since lock state is global, maybe this is better done in the message at the top of the page that we plan to add?

Doesn't hurt to have it here too. We can always adjust later.

src/components/Pill/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+        break;
+      case 'published':
+        pills.push({
+          text: 'Live',
```
mikix: Huh, I looked that up and that _is_ Carol's wording. I'm a little surprised that we'd use that for partners. Feels techy to me.

src/components/EditCoursePage/index.jsx
```diff
@@ -154,8 +154,21 @@ class EditCoursePage extends React.Component {
...
+    const coursePublished = course_runs && course_runs.some(courseRun =>
+      PUBLISHED === courseRun.status) && PUBLISHED;
+    courseStatuses.push(coursePublished);
+    const courseReviewed = course_runs && !coursePublished && course_runs.some(courseRun =>
+      REVIEWED === courseRun.status) && REVIEWED;
+    courseStatuses.push(courseReviewed);
+    const courseUnpublished = course_runs && !coursePublished && !courseInReview &&
+      !courseReviewed && course_runs.some(courseRun => UNPUBLISHED === courseRun.status) &&
+      UNPUBLISHED;
+    courseStatuses.push(courseUnpublished);
```
mikix: These statuses aren't mutually exclusive? Why the list pushing instead of an if/else situation?

mikix:
About the look of it - I like it. I might prefer if the pills were vertically aligned? Probably right-aligning them would be easiest way to do that. But a matter of taste. And I realize her slides have them like you do - but I thin that's partly because she has right-aligned the submit buttons inline which we never ended up doing (and probably won't?) - so we have space there that she didn't when making the slides.

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
```diff
@@ -153,7 +160,6 @@ export const BaseCollapsibleCourseRunFields = ({
               *  through the handler and redux-form's submit if it passes validation. If validation
               *  fails, report the form issues back without triggering submission.
               */
-              // We manually check the form for validity, and then pass the targeted
```
bbaker6225: y tho? 😢

Dillon-Dumesnil: Because I had an AC of common sense

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
Dillon-Dumesnil: Starting a conversation in the fedx channel about this

src/components/EditCoursePage/EditCourseForm.jsx
Dillon-Dumesnil: This is Dillon UI that was somewhat discussed over slack.

src/components/EditCoursePage/EditCourseForm.jsx
Dillon-Dumesnil: Yeah bud

src/components/EditCoursePage/index.jsx
Dillon-Dumesnil: They are not mutually exclusive for a Course. One scenario is the course is live and has a different course run in review so it has both the live and in review statuses. Eventually, we will also support having an `unsaved edits` status (or something similar) that will also add to this list.

src/components/Pill/index.jsx
Dillon-Dumesnil: I agree with that and will leave a comment in the ticket that handles the page level status messages.

I think it's nice to still leave in here so you know which course run(s) are causing you to be in review.

src/components/Pill/index.jsx
Dillon-Dumesnil: It allows bool here because `courseStatuses` come in as either one of the switch cases or False if the condition to show it is not true.

mikix: Ah, I see now. Why not just have those switches do something like:?
```
if (condition)
   push(status)
```

Dillon-Dumesnil: Because the conditions are only checked as part of adding a pill to a course because we have to check what the status of the course runs are.
So the switches are only looking at the values of possible CourseRun statuses, but https://github.com/edx/publisher-frontend/pull/121/files#diff-fa7269192ac8344c05273e80552d7c0eR161 is setting which statuses should be applied to the course by doing multiple checks and if the check is not valid, it returns `false`

Dillon-Dumesnil: Updated to be if statements.

PR #124, opened 2019-04-22
Fix props warning on pills test

bbaker6225:
Fixes Screen Shot 2019-04-22 at 1 49 48 PM


Dillon-Dumesnil:
My bad

bbaker6225:
@Dillon-Dumesnil I owed you after `// We manually check the form for validity, and then pass the targeted`

PR #126, opened 2019-04-22
Add required for submission label to courses/runs

bbaker6225:
DISCO-828

Actual validation to come in a separate PR

### DIFF
--- a/src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
+++ b/src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
@@ -8,6 +8,7 @@ import { connect } from 'react-redux';
 import ButtonToolbar from '../ButtonToolbar';
 import FieldLabel from '../FieldLabel';
 import Pill from '../Pill';
+import { PUBLISHED } from '../../data/constants';
 import RenderInputTextField from '../RenderInputTextField';
 import RenderSelectField from '../RenderSelectField';
 import StaffList from '../StaffList';
@@ -80,7 +81,7 @@ export const BaseCollapsibleCourseRunFields = ({
           type="date"
           component={RenderInputTextField}
           format={value => getDateString(value)}
-          label={<FieldLabel text="Start date" required />}
+          label={<FieldLabel text="Start date" required requiredForSubmit />}
           placeholder="mm/dd/yyyy"
           required
           disabled={courseInReview}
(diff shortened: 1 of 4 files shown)


src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
```diff
@@ -169,7 +170,7 @@ export const BaseCollapsibleCourseRunFields = ({
               }
             }}
           >
-            {courseRuns[index].status === 'published' ? (
+            {courseRuns[index].status === PUBLISHED ? (
```
bbaker6225: 🤷‍♂️ why not use the constant if we have it

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
```diff
@@ -80,7 +81,7 @@ export const BaseCollapsibleCourseRunFields = ({
           type="date"
           component={RenderInputTextField}
           format={value => getDateString(value)}
-          label={<FieldLabel text="Start date" required />}
+          label={<FieldLabel text="Start date" required requiredForSubmit />}
```
bbaker6225: In my mind it's nice to have both in case `required` is one day removed but it's still `requiredForSubmit`... but happy to remove it if there is disagreement

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -112,42 +112,42 @@ export class BaseEditCourseForm extends React.Component {
...
               component={RenderInputTextField}
               type="text"
-              label={<FieldLabel text="Title" required />}
+              label={<FieldLabel text="Title" required requiredForSubmit />}
               required
               disabled={courseInReview}
             />
             <div>
-              <FieldLabel id="number" text="Number" className="mb-2" />
+              <FieldLabel id="number" text="Number" className="mb-2" requiredForSubmit />
```
bbaker6225: Seems weird to me to have `requiredForSubmit` on a field you can't edit... but old publisher does it so why not

bbaker6225:
screencapture-localhost-18400-courses-8dc363cd-d7b5-4251-b6d9-23d316bec3c6-edit-2019-04-22-15_09_47

src/components/EditCoursePage/EditCourseForm.jsx
mikix: I actually hate that old publisher does it. I vote dropping it.

src/components/EditCoursePage/CollapsibleCourseRunFields.jsx
mikix: Naw, I think that's a fair call.

PR #127, opened 2019-04-23
DISCO-794: Error messaging update

jlajoie:


src/data/actions/courseOptions.js
```diff
@@ -38,15 +40,15 @@ function fetchCourseOptions(id) {
 
         // Confirm it looks vaguely correct
         if (!course || !('actions' in course)) {
-          throw Error('Did not understand response.');
```
Dillon-Dumesnil: Why delete this poor period?

jlajoie: Adding it in later because our API error messages don't have periods, and want it to be consistent without saying `Some full error string: Did not understand response..`

jlajoie:
Story for tests created at https://openedx.atlassian.net/browse/DISCO-843

PR #129, opened 2019-04-24
Allow cents for price and use fancy submit buttons

mikix:
- Bumps paragon to 4.1.4
- Drops LoadingMessage and TableLoadingOverlay
- Drops _variables.scss in favor of direct bootstrap
- Adds ActionButton, which is a thin wrapper around StatefulButton


src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -96,6 +97,15 @@ export class BaseEditCourseForm extends React.Component {
     const languageOptions = (courseRunOptions &&
       this.parseOptions(courseRunOptions.content_language.choices));
 
+    let submitState = 'default';
+    if (submitting) {
+      submitState = 'pending';
+    } else if (pristine) {
+      // FIXME: Once this form is correctly reset to a pristine state after a successful submit,
```
Dillon-Dumesnil: We will need to manually figure out how to do this given the `keepDirtyOnReinitialize` form parameter defined at the bottom of this file.

PR #133, opened 2019-04-24
Update to use new DRF endpoint

mikedikan:
DISCO-833

### DIFF
--- a/src/components/StaffList/StaffList.test.jsx
+++ b/src/components/StaffList/StaffList.test.jsx
@@ -36,22 +36,16 @@ const owners = [{ key: 'MITx' }];
 const autoCompletePersonResponses = {
   long: [
     {
-      id: '5',
-      text: {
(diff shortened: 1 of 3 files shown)


src/components/StaffList/StaffList.test.jsx
```diff
@@ -106,19 +100,18 @@ describe('StaffList', () => {
   });
 
   it('gets/clears suggestions for autocomplete', (done) => {
-    mockClient.onGet('http://localhost:18381/admin/course_metadata/person-autocomplete/?q=long&serialize=1&org=MITx')
-      .replyOnce(200, JSON.stringify({
-        results: autoCompletePersonResponses.long,
-      }));
+    //                http://localhost:18381/api/v1/search/person_typeahead/?q=long&org=MITx
```
dianakhuang: I don't think we need this comment anymore?

PR #135, opened 2019-04-25
Fix key warning on pills

bbaker6225:


src/components/Pill/index.jsx
```diff
@@ -40,7 +40,7 @@ const Pill = ({ statuses }) => {
     }
   });
   return pills.map(pill => (
-    <span className={classNames('ml-2', pill.className)}>
+    <span className={classNames('ml-2', pill.className)} key={pill.text}>
```
Dillon-Dumesnil: If I'm understanding this correctly, I don't think this will ensure uniqueness unfortunately since you can have multiple course runs in the same state.

bbaker6225: They just need to be unique in the iterator, and the text shouldn't be the same for multiple pills on the same collapsible right?

Dillon-Dumesnil: Ah okay, I misunderstood this uniqueness constraint. Yes, each collapsible should have unique pills.

PR #136, opened 2019-04-25
Add tooltip support

mikix:
This adds a single tooltip, as a base for further work.

DISCO-746

Screenshot from 2019-04-25 11-15-55

Screenshot from 2019-04-25 11-16-02


src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -117,12 +117,44 @@ export class BaseEditCourseForm extends React.Component {
           <Collapsible
             title={this.formatCourseTitle(title, courseStatuses)}
             key="Test Key"
+            isOpen
```
mikix: I just threw this unrelated change in here - it came out of bug bush.

package.json
```diff
@@ -30,6 +30,7 @@
     "eslint": "^5.16.0",
     "font-awesome": "^4.7.0",
     "history": "^4.7.2",
+    "jsx-to-string": "^1.4.0",
```
mikix: This and react-tooltip are used by prospectus already to do this.

src/components/FieldHelp/index.jsx
```diff
@@ -0,0 +1,83 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import jsxToString from 'jsx-to-string';
+import ReactTooltip from 'react-tooltip';
+
+
+class FieldHelp extends React.Component {
```
mikix: This class stolen from prospetus almost verbatem.

mikix:
Oh, tests would actually be appriopriate for this - there is user interaction and all here. I'll add those

src/components/EditCoursePage/EditCourseForm.jsx
Dillon-Dumesnil: Looks like "Test Key" was originally put in by Jeff and I believe it has to do with it being in a Collapsible and requiring a unique key value. Not sure why it's that and could probably be switched to a uuid or something, but @jlajoie or @bbaker6225  can probably speak better to it.

src/components/FieldLabel/index.jsx
```diff
@@ -10,21 +11,26 @@ const FieldLabel = (props) => {
...
+      {props.helpText && <FieldHelp id={`${props.id}-help`} tip={props.helpText} />}
       {requireText}
     </div>
   );
 };
 
 FieldLabel.defaultProps = {
   className: '',
+  helpText: '',
+  id: 'label', // will not be suitable for production, but useful default for tests
```
Dillon-Dumesnil: Why not just make it a required field?

mikix: Because I was being _laaaaazy_. I can fix.

mikix: Oh actually - because we haven't converted all the field labels to have helpText yet - and the id is only required if you pass helpText.

So maybe I'll just set the default to null and only show helpText if it's set.

src/components/EditCoursePage/EditCourseForm.jsx
jlajoie: Yeah that shouldn't be this, just make it something unique, sorry for the mistake!

bbaker6225: You shouldn't need `key` at all since it's not in an iterator

PR #137, opened 2019-04-25
Mark different fields as required depending on submit source

bbaker6225:
This doesn't do any of the custom validation or focus management stuff, just triggers which fields are required or not based off what triggered the edit course form submission


mikix:
I didn't test, but assume it's great.

PR #145, opened 2019-04-30
More small fixes

mikix:
- Clean up URLs a bit (mostly dropping /edit)
- Stop the image field from blocking submit-for-review by making it
  non required (we should revisit this later)
- Prevent an 'undefined' staffer from being added if you cancel making
  a new staffer
- Prevent the new staffer form from being filled out with a previous
  staffer you were editing by resetting it on load

Filed https://openedx.atlassian.net/browse/DISCO-883 about the image issue.


src/components/CourseTable/index.jsx
```diff
@@ -32,7 +32,7 @@ const CourseTable = () => {
 
   const formatCourseData = courses => courses.map(course => ({
     ...course,
-    title: (<Link to={`/courses/${course.uuid}/edit/`}>{course.title}</Link>),
+    title: (<Link to={`/courses/${course.uuid}`}>{course.title}</Link>),
```
Dillon-Dumesnil: Thoughts on trailing slash or not?

mikix: I'm not a huge fan? But could be persuaded otherwise.

But if we add them, I'd like to be consistent and add them everywhere. So `/instructors/new/`, etc. We were a mishmash before.

Dillon-Dumesnil: Definitely agree with consistency, but I personally don't have strong feelings one way or the other.

I believe without the `exact` it can match regardless of the trailing slash so is that potentially the preferred way and internally we remove the slashes everywhere? I remember being annoyed on Discovery when I tried to go to a url that 404'd or something because I forgot the trailing slash and it took me a few minutes to figure out what was happening

mikix: Even with `exact`, it will match with or without trailing slash.

mikix: I believe `exact` just prevents <uuid>/blarg from matching.

Dillon-Dumesnil: That feels like common sense. Weird that you have to specify that and the default is false.

All good then! And we just need to make sure we start using no trailing slash everywhere

mikix: Agreed. I don't get it.

PR #147, opened 2019-04-30
Add tooltips when creating a course or a run

mikix:
DISCO-745


src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -20,7 +23,7 @@ class BaseCreateCourseForm extends React.Component {
   }
 
   processOrganizations(organizations) {
-    const orgSelectList = [{ label: 'Select user organization', value: '' }];
+    const orgSelectList = [{ label: 'Select organization', value: '' }];
```
mikix: Unrelated change - just that course teams are going to be confused by "user" -- the learner or them? Figured, just drop it.

PR #149, opened 2019-05-01
Updating the frontend to send an indicator variable for when we should

Dillon-Dumesnil:
update the course after it has gone live. Fixing price on create and edit to detect and send changes.

Also allowing updates to mode and price until a course run is in review. Price will be hidden unless the mode is professional or verified.

### DIFF
--- a/src/components/CreateCoursePage/CreateCourseForm.jsx
+++ b/src/components/CreateCoursePage/CreateCourseForm.jsx
@@ -8,7 +8,12 @@ import ActionButton from '../ActionButton';
 import ButtonToolbar from '../ButtonToolbar';
 import FieldLabel from '../FieldLabel';
 
-import { AUDIT_TRACK, VERIFIED_TRACK, PROFESSIONAL_TRACK } from '../../data/constants';
+import {
+  AUDIT_TRACK,
+  ENTITLEMENT_TRACKS,
+  PROFESSIONAL_TRACK,
+  VERIFIED_TRACK,
+} from '../../data/constants';
 import { endDateHelp, enrollmentHelp, startDateHelp, titleHelp } from '../../helpText';
 
 
@@ -35,6 +40,7 @@ class BaseCreateCourseForm extends React.Component {
 
   render() {
     const {
+      currentValues,
       handleSubmit,
       organizations,
       pristine,
@@ -119,17 +125,17 @@ class BaseCreateCourseForm extends React.Component {
             }
             required
           />
-          <Field
+          {ENTITLEMENT_TRACKS.includes(currentValues.enrollmentTrack) && <Field
             name="price"
             component={RenderInputTextField}
-            input={{
+            extraInput={{
               min: 0.01,
               step: 0.01,
             }}
             type="number"
             label={<FieldLabel text="Price (USD)" required />}
             required
-          />
+          />}
           <h2>First run of your Course</h2>
           <hr />
           <Field
@@ -179,6 +185,7 @@ class BaseCreateCourseForm extends React.Component {
 BaseCreateCourseForm.defaultProps = {
   isCreating: false,
   pristine: true,
+  currentValues: {},
 };
 
 BaseCreateCourseForm.propTypes = {
@@ -192,6 +199,7 @@ BaseCreateCourseForm.propTypes = {
     start: PropTypes.string,
     end: PropTypes.string,
   }).isRequired,
+  currentValues: PropTypes.shape({}),
   organizations: PropTypes.arrayOf(PropTypes.shape({
     name: PropTypes.string.isRequired,
     key: PropTypes.string.isRequired,
--- a/src/components/CreateCoursePage/CreateCourseForm.test.jsx
+++ b/src/components/CreateCoursePage/CreateCourseForm.test.jsx
@@ -8,6 +8,14 @@ const organizations = [
 ];
 
 describe('CreateCourseForm', () => {
+  const initialValues = {
+    org: 'edx',
+    title: 'Hello',
+    number: 'edx101',
+    enrollmentTrack: 'verified',
+    price: 100.00,
+  };
+
   it('renders html correctly with no orgs', () => {
     const component = shallow(<BaseCreateCourseForm
       handleSubmit={() => {}}
@@ -16,20 +24,17 @@ describe('CreateCourseForm', () => {
     />);
     expect(component).toMatchSnapshot();
   });
+
   it('renders html correctly with data', () => {
     const component = shallow(<BaseCreateCourseForm
       handleSubmit={() => {}}
-      initialValues={{
-        org: 'edx',
-        title: 'Hello',
-        number: 'edx101',
-        enrollmentTrack: 'verified',
-        price: 100.00,
-      }}
+      initialValues={initialValues}
+      currentValues={initialValues}
       organizations={organizations}
     />);
     expect(component).toMatchSnapshot();
   });
+
   it('renders html correctly while submitting', () => {
     const component = shallow(<BaseCreateCourseForm
       submitting
@@ -37,6 +42,7 @@ describe('CreateCourseForm', () => {
       pristine={false}
       handleSubmit={() => {}}
       initialValues={{}}
+      currentValues={initialValues}
       organizations={organizations}
     />);
     expect(component).toMatchSnapshot();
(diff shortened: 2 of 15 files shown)


Dillon-Dumesnil:
What you can't see in here and will come in the course-discovery PR (https://github.com/edx/course-discovery/pull/1884), when a course with an Audit mode selected is Reviewed, I'm going to save a fake SKU into the draft entitlement sku field which will maintain the disable if the course goes from Reviewed back into Incomplete.

mikix:
You mention a fake SKU - I believe the intention is that we actually create products upon first review success. So could just use the real SKU. Although I guess you mean currently, where we don't actually make them. But we want to fake that flow. Gotcha.

src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -179,6 +185,7 @@ class BaseCreateCourseForm extends React.Component {
 BaseCreateCourseForm.defaultProps = {
   isCreating: false,
   pristine: true,
+  currentValues: undefined,
```
mikix: I don't think `undefined` qualifies as a shape, which is how you define this prop. If you defaulted to `{}`, you could also probably drop the `currentValues &&` check above?

src/components/CreateCoursePage/index.jsx
```diff
@@ -110,6 +112,7 @@ CreateCoursePage.defaultProps = {
   fetchOrganizations: () => {},
   courseInfo: {},
   createCourse: () => {},
+  formValues: undefined,
```
mikix: Same.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -653,37 +659,33 @@ export class BaseEditCourseForm extends React.Component {
...
+                  id="mode.label"
+                  text="Enrollment track"
+                  helpText={enrollmentHelp}
                 />
-              </React.Fragment>
-            )}
+              }
+              extraInput={{ onInvalid: this.openCollapsible }}
+              options={this.getEnrollmentTrackOptions()}
+              disabled={courseInReview || (!!entitlement && !!entitlement.sku)}
```
mikix: Taking a hard stance in favor of `!!`, eh?

src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -119,17 +125,17 @@ class BaseCreateCourseForm extends React.Component {
             }
             required
           />
-          <Field
+          {currentValues && ENTITLEMENT_TRACKS.includes(currentValues.enrollmentTrack) && <Field
```
mikix: Feels weird to go up several layers for the currentValues from fromValues etc, just to grab the value of the field next to you here. But I suppose that's correct.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -653,37 +659,33 @@ export class BaseEditCourseForm extends React.Component {
...
+            />
+            {currentValues && ENTITLEMENT_TRACKS.includes(currentValues.mode) && <Field
```
mikix: If this field doesn't exist because of audit mode, when we PATCH, do we serialize anything weird for price?

src/components/EditCoursePage/index.jsx
```diff
@@ -53,6 +53,9 @@ class EditCoursePage extends React.Component {
       // If a course run triggered the submission, mark it as not a draft
       const submittedRun = courseRuns.find(run => run.key === targetRun.key);
       submittedRun.draft = false;
+      // If we are updating a published course run, we need to also update the course.
+      // We want to use the same indicator of draft = false for consistency.
+      courseData.draft = !(submittedRun.status === PUBLISHED);
```
mikix: This would read cleaner to me as `courseData.draft = submittedRun.status !== PUBLISHED`

Or honestly, most readable as:
```
if submittedRun.status === PUBLISHED:
    courseData.draft = False
```

Just because setting False to *do* a thing feels weird. And this makes the *doing* a little more explicit.

src/components/EditCoursePage/index.jsx
```diff
@@ -53,6 +53,9 @@ class EditCoursePage extends React.Component {
       // If a course run triggered the submission, mark it as not a draft
       const submittedRun = courseRuns.find(run => run.key === targetRun.key);
       submittedRun.draft = false;
+      // If we are updating a published course run, we need to also update the course.
```
mikix: I might say `we need to also publish the course`

Dillon-Dumesnil:
So I actually hadn't thought about faking the flow for Usability testing, and am not actually sure if that's the best route to take. I would have to think through any implications of that first.

What I'm referring to here is that if you have an Audit course, we do not create an Audit product in Ecommerce on first review. That only happens for course modes that have a price (Verified and Professional). Since we are using the SKU as an indicator to disable changing the mode, I am planning on saving a fake SKU into the **draft** Audit course entitlement (you're correct that the non-audit entitlements will get these from ecommerce) so even if you select Audit as your course mode, we will still disable changing the mode.

Does that make sense?

mikix:
I see. A little hacky maybe. But that's because what we really want is a serialized field like "has ever been reviewed" and we're looking for side effects.

I'd say fine for now - but we might want to consider adding a new status like REVIEWED_BUT_EDITED (or whatever) to replace kicking back to UNPUBLISHED.

src/components/EditCoursePage/EditCourseForm.jsx
Dillon-Dumesnil: I thought about not having it as `!!`, but disabled is expecting a boolean and that's the way of casting to bool. Otherwise I was running into scenarios where `entitlement = {}` and `entitlement.sku` evaluated to `null` so `disabled={null}` which felt worse.

src/components/CreateCoursePage/CreateCourseForm.jsx
Dillon-Dumesnil: formValues is the way to get the **live** updates from the form. It definitely does take a couple layers to get here though. I wish it were simpler. https://redux-form.com/8.1.0/docs/api/selectors.md/

Dillon-Dumesnil:
We are going to speak to Carol about this. If it is possible to upgrade from Audit to Verified or Professional, we can simplify some of the logic and leave mode editable as long as it is Audit.

src/components/CreateCoursePage/CreateCourseForm.test.jsx
```diff
@@ -8,6 +8,14 @@ const organizations = [
 ];
 
 describe('CreateCourseForm', () => {
+  const initialValues = {
+    org: 'edx',
```
bbaker6225: capital x

src/components/CreateCoursePage/CreateCourseForm.test.jsx
```diff
@@ -8,6 +8,14 @@ const organizations = [
...
+    title: 'Hello',
+    number: 'edx101',
```
bbaker6225: capital x

src/components/EditCoursePage/EditCourseForm.jsx
Dillon-Dumesnil: On creation, we will send `price: undefined` from the frontend and the backend gives a default value of 0.00 if price does not exist. So in the end, we will create a draft entitlement with mode Audit and price 0.00. Any future saves, will send price = 0.00.

So it's documented, if you start verified and set a price of say $50 and then later switch to Audit. The Audit saves will still be sent with the $50 because that was the most recent saved form value so it still gets sent. This will update the draft entitlement to be Audit with a price, but we still won't create an official entitlement because the mode is not valid. The silver lining of this is that if the user is switching between audit and verified, we are "saving" their last known price value.

src/components/CreateCoursePage/index.jsx
```diff
@@ -135,6 +138,7 @@ CreateCoursePage.propTypes = {
   }),
   fetchOrganizations: PropTypes.func,
   createCourse: PropTypes.func,
+  formValues: PropTypes.shape({}),
```
jlajoie: Do we care to structure this more than just any object?

Dillon-Dumesnil: I didn't end up bothering because we use the same situation on the EditCoursePage/Form and to list all the fields there felt like a lot. It's doable here if you think there's value in it because it is only 5 or so fields, but I figured just declaring it as an object would be good

PR #150, opened 2019-05-01
Open collapsible when a validation error occurs inside

mikix:
- Rename CollapsibleCourseRunFields -> CollapsibleCourseRuns
- Add CollapsibleCourseRun component to hold one single run
- Add isInvalid handlers for fields that will open the collapsible

Not all fields currently report isInvalid - that's a work in progress.

DISCO-882


src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -0,0 +1,367 @@
...
+
+    this.state = {
+      open: false,
+    };
+
+    this.openCollapsible = this.openCollapsible.bind(this);
+    this.setCollapsible = this.setCollapsible.bind(this);
+  }
+
+  setCollapsible(open) {
```
Dillon-Dumesnil: Is setCollapsible a thing because we want to stay open to the idea of being able to pass in false?

Why not just have openCollapsible and closeCollapsible?

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -120,7 +141,8 @@ export class BaseEditCourseForm extends React.Component {
           <Collapsible
             title={this.formatCourseTitle(title, courseStatuses)}
             key="Test Key"
-            isOpen
+            isOpen={this.state.open}
```
Dillon-Dumesnil: Up to you, but I might deconstruct this. If we are only using the state in this one spot and with one variable, not totally necessary though.

src/components/EditCoursePage/CollapsibleCourseRuns.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import CollapsibleCourseRun from './CollapsibleCourseRun';
+
+const CollapsibleCourseRuns = ({
+  courseRuns,
+  fields,
+  ...others
```
Dillon-Dumesnil: I think we should have a quick team discussion to standardize how we want to name props that are just meant to be passed down. I've seen `others`, `extraInput`, and `passedProps` all in this PR

src/components/EditCoursePage/CollapsibleCourseRun.jsx
mikix: It's written that way because it's easier to have it just be the onToggle handler - which takes `false` or `true` - we are trying to stay in sync with the internal state of the collapsible.

src/components/EditCoursePage/CollapsibleCourseRuns.jsx
mikix: `extraInput` and `passedProps` are both shapes themselves with specific targets in mind. Like `extraInput` is aimed at the underlying `input` element.

So those I could see being looser about naming - they have their own intent they are trying to convey.

But in this case, we are looking for something that means "the rest of the props". Would `otherProps` be nicer? `props` by itself feels too ambiguous.

Dillon-Dumesnil: I still feel like at the end of the day, both extraInput and passedProps are props we are handing down the chain, but I do see your point.

Agreed that props by itself is too ambiguous. `otherProps` is nicer, but I don't know. Does FedX have any thoughts on this?

https://reactjs.org/docs/higher-order-components.html#convention-pass-unrelated-props-through-to-the-wrapped-component Calls is `passThroughProps`, but I'm not in love with that either

mikix: `passThroughProps` is probably just an example. But it is in the official docs on how to do this. And it is explicit. I'm fine with it.

Dillon-Dumesnil: Blargh. I will accept any executive decision made.

mikix: I just used `passThroughProps` and fixed a few other instances of other names.

mikix: I think the important thing is consistency - easier to find all instances and rename them in future, if they all use `passThroughProps` for now.

PR #154, opened 2019-05-03
Client side validation/focus for non-input fields

bbaker6225:
DISCO-878

Note: we don't pass through props on fields within field arrays which rely on `meta` to avoid "polluting" the field meta info with higher up meta information

### DIFF
--- a/src/components/EditCoursePage/CollapsibleCourseRun.jsx
+++ b/src/components/EditCoursePage/CollapsibleCourseRun.jsx
@@ -309,7 +309,6 @@ class CollapsibleCourseRun extends React.Component {
             </div>
           }
         />
-        {/* We don't use pass through props on staff to avoid wiping out meta info */}
         <Field
           name={`${courseId}.staff`}
           component={StaffList}
--- a/src/components/EditCoursePage/CollapsibleCourseRun.jsx
+++ b/src/components/EditCoursePage/CollapsibleCourseRun.jsx
@@ -6,11 +6,13 @@ import { Collapsible } from '@edx/paragon';
 
 import ActionButton from '../ActionButton';
 import ButtonToolbar from '../ButtonToolbar';
+import courseSubmitInfo from '../../data/actions/courseSubmitInfo';
 import FieldLabel from '../FieldLabel';
 import Pill from '../Pill';
 import RenderInputTextField from '../RenderInputTextField';
 import RenderSelectField from '../RenderSelectField';
 import StaffList from '../StaffList';
+import store from '../../data/store';
 import TranscriptLanguage from './TranscriptLanguage';
 
 import { IN_REVIEW_STATUS, PUBLISHED } from '../../data/constants';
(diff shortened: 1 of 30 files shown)


src/utils/validation.js
```diff
@@ -0,0 +1,123 @@
...
+ *         staff: "Required"
+ *       }
+ *     ]
+ *   };
+ *   The param will be null if there are no errors.
+ *
+ * @returns {string} - The first field name with a validation error
+ */
+const getFieldName = (errors) => {
+  if (errors) {
```
mikix: optional nit: I usually write this kind of function inverse style with an early exit. So `if (!errors) return null;`  That way you don't need an extra layer of indenting for the real logic.

src/utils/validation.js
```diff
@@ -0,0 +1,123 @@
...
+      }
+    });
+
+    // Validate all the fields required for submission in the submitting course run
+    errors.course_runs = [];
+    values.course_runs.forEach((run) => {
+      const { key: targetKey } = targetRun;
+      const isSubmittingRun = run.key === targetKey;
+      if (isSubmittingRun) {
+        const runRequiredFields = ['transcript_languages', 'staff'];
```
mikix: Is there a way to check for like, a field attribute or something, rather than hardcoding? Hardcoding is not the worst. Not a thing that changes often. But just something I bumped on.

But I also see that languages below also requires special logic. Maybe hardcoding is the way to go.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -771,6 +770,22 @@ const EditCourseForm = compose(
...
+      if (
+        props.submitting ||
+        (nextProps && nextProps.submitting) ||
+        props.targetRun ||
+        (nextProps && nextProps.targetRun)
+      ) {
+        return true;
+      }
+
+      return false;
```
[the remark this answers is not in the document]
bbaker6225: yes!

src/utils/validation.js
bbaker6225: smart smart

src/utils/validation.js
bbaker6225: unfortunately values comes back as a giant object of every single field name and its value, so it seems like we have to hardcode unless we're validating all fields in this way

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -140,7 +143,7 @@ export class BaseEditCourseForm extends React.Component {
 
     return (
       <div className="edit-course-form">
-        <form id={id}>
+        <form id={id} onSubmit={handleSubmit}>
```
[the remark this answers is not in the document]
bbaker6225: I thought that I wouldn't be able to pass the `targetRun` up the submit chain without overriding the built in `handleSubmit`... but I was wrong

src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -307,32 +309,29 @@ class CollapsibleCourseRun extends React.Component {
             </div>
           }
         />
+        {/* We don't use pass through props on staff to avoid wiping out meta info */}
```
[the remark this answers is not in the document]
bbaker6225: 👍

PR #155, opened 2019-05-03
Aj/disco 888

staubina:
Screen Shot 2019-05-03 at 4 17 06 PM

### DIFF
--- a/src/components/EditCoursePage/EditCourseForm.jsx
+++ b/src/components/EditCoursePage/EditCourseForm.jsx
@@ -4,7 +4,7 @@ import { Field, FieldArray, reduxForm, stopSubmit } from 'redux-form';
 import { Link } from 'react-router-dom';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
-import { Collapsible } from '@edx/paragon';
+import { Collapsible, Icon } from '@edx/paragon';
 
 import ActionButton from '../../components/ActionButton';
 import ButtonToolbar from '../../components/ButtonToolbar';
@@ -63,6 +63,35 @@ export class BaseEditCourseForm extends React.Component {
     }
 
     return data.actions.PUT;
+  }
+
+  getAddCourseRunButton(courseInReview, pristine, uuid) {
+    /** Disabling a Link is discouraged and disabling a button within a link results
+     * in a Disabled button with a link that will still underline on hover.
+     * This method will remove the Link when disabling the button.
+     */
+    if (courseInReview) {
+      return '';
+    }
+
+    const courseRunButton = (
+      <button className="btn btn-block rounded mt-3 new-run-button" disabled={!pristine}>
+        <Icon className="fa fa-plus" /> Add Course Run
+      </button>);
+
+    let buttonWrapper = (
+      <Link to={`/courses/${uuid}/rerun`}>
+        {courseRunButton}
+      </Link>);
+
+    if (!pristine) {
+      buttonWrapper = (
+        <React.Fragment>
+          {courseRunButton}
+        </React.Fragment>
+      );
+    }
+    return buttonWrapper;
   }
 
   getCourseRunOptions() {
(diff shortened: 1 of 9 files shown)


src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -696,22 +696,25 @@ export class BaseEditCourseForm extends React.Component {
...
-            </Link>
             <ActionButton
               disabled={courseInReview || submitting}
               labels={{
-                default: 'Save Course',
+                default: 'Save & Continue Editing',
                 pending: 'Saving Course',
                 complete: 'Course Saved',
               }}
+              className="btn-outline-primary"
```
Dillon-Dumesnil: I don't think you need to pass this in when you already have your `primary` indicator variable that sets this class when `primary = false`

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -696,22 +696,25 @@ export class BaseEditCourseForm extends React.Component {
             courseSubmitting={submitting}
             {...this.props}
           />
+          <Link to={`/courses/${uuid}/rerun`}>
+            <button
+              id="new-run-button"
```
Dillon-Dumesnil: Nit: I might check with Alasdair, but I think using ids for styling is typically frowned upon and it's better to use a class.

src/components/EditCoursePage/EditCourseForm.jsx
staubina: Good catch forgot I was still passing this.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -696,22 +696,32 @@ export class BaseEditCourseForm extends React.Component {
...
+          {(!courseInReview && pristine &&
+          <Link to={`/courses/${uuid}/rerun`}>
+            <button
+              className="btn btn-block rounded mt-3 new-run-button"
+            >
+              <Icon className="fa fa-plus" /> Add Course Run
+            </button>
+          </Link>
+          )}
+          {(!pristine &&
```
Dillon-Dumesnil: Should this also have the `!courseInReview` check? Otherwise we will show the disabled button when the course is in review if the form is not pristine. Or is it safe because we disable editing the form once it is in review so it should remain pristine?

Might have weird original behaviour when you first submit for review because that will change `courseInReview` to true so the working link won't be displayed, but since we don't make the form pristine again, the disabled button will still show and then when they refresh the page, it will clear back to not showing anything.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -696,22 +696,32 @@ export class BaseEditCourseForm extends React.Component {
...
+            <button
```
Dillon-Dumesnil: This is also a lot of repeated code between these two states. Maybe you can make a variable like `disableAddRunButton` and set it equal to `!pristine` and then just have one button with `disabled={disableAddRunButton}`

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -728,6 +751,7 @@ export class BaseEditCourseForm extends React.Component {
   }
 }
 
+
```
Dillon-Dumesnil: Nit: remove this

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -65,6 +65,35 @@ export class BaseEditCourseForm extends React.Component {
     return data.actions.PUT;
   }
 
+  getAddCourseRunButton(courseInReview, pristine, uuid) {
+    /** Disabling a Link is discouraged and disabling a button within a link results
+     * in a Disabled button with a link that will still underline on hover.
```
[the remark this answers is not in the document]
staubina: It is not the preferred method and is discouraged.

src/components/EditCoursePage/EditCourseForm.jsx
staubina: Doing this then squashing

PR #165, opened 2019-05-13
Updated the handleCourseEdit to use a copy of courseData, not original

staubina:


src/components/EditCoursePage/index.jsx
```diff
@@ -89,32 +88,34 @@ class EditCoursePage extends React.Component {
...
+    updatedCourseEditData.video = { src: updatedCourseEditData.videoSrc };
+    updatedCourseEditData.key = key;
+    updatedCourseEditData.uuid = uuid;
+    return updatedCourseEditData;
   }
 
   handleCourseEdit(courseData) {
     /*
-      Need to do some preprocessing before sending anything to course-discovery.
+      Need to do some pre-processing before sending anything to course-discovery.
```
Dillon-Dumesnil: Got to love that Grammerly

src/components/EditCoursePage/index.jsx
```diff
@@ -78,8 +78,7 @@ class EditCoursePage extends React.Component {
     /* eslint-enable no-param-reassign */
   }
 
-  prepareCourse(courseData) {
-    /* eslint-disable no-param-reassign */
+  prepareCourse(courseEditData) {
```
Dillon-Dumesnil: Nit: Keep the input parameter as courseData since that is what we are passing in.

Just felt a little weird to me when the input parameter of the function is courseEditData and the value to assign the function to is courseEditData (line 139)

src/components/EditCoursePage/index.jsx
```diff
@@ -134,11 +136,9 @@ class EditCoursePage extends React.Component {
 
     // Process course run info
     this.prepareCourseRuns(modifiedCourseRuns, courseData);
+    const courseEditData = this.prepareCourse(courseData);
 
-    // Process course info
```
Dillon-Dumesnil: Nit: Keep this comment since we have the course run comment above

src/components/EditCoursePage/index.jsx
```diff
@@ -126,6 +127,7 @@ class EditCoursePage extends React.Component {
     const {
       editCourse,
     } = this.props;
+
     // Create a deep copy of course runs since we modify their properties
     let modifiedCourseRuns = jsonDeepCopy(courseData.course_runs);
```
Dillon-Dumesnil: I don't have strong preferences on which way you choose, but I'd prefer consistency between prepareCourseRuns and prepareCourse. Like we are doing the jsonDeepCopy outside and then letting prepareCourseRuns modify the modifiedCourseRuns, but we do the opposite in prepareCourse.

src/components/EditCoursePage/index.jsx
staubina: This one was my IDE :D

src/components/EditCoursePage/index.jsx
staubina: @Dillon-Dumesnil I made updates for this request, please review when you have time.

src/components/EditCoursePage/index.jsx
```diff
@@ -126,19 +125,31 @@ class EditCoursePage extends React.Component {
...
-    // Don't send any courses in review - backend will reject them
-    modifiedCourseRuns = modifiedCourseRuns.filter(run => !IN_REVIEW_STATUS.includes(run.status));
+    const { targetRun } = this.state;
 
-    // Process course run info
-    this.prepareCourseRuns(modifiedCourseRuns, courseData);
+    // Process course run info from courseData
+    const modifiedCourseRuns = this.prepareCourseRuns(courseData);
+    // Update the Draft status based on run status
+    let courseEditData = jsonDeepCopy(courseData); // eslint-disable-line prefer-const
```
Dillon-Dumesnil: Thinking about it more, maybe we let `prepareCourseRuns` and `prepareCourse` modify their input data since they are merely helper functions to `handleCourseEdit`. Then we can just move this line to the top of the function and remove the other deepcopys.

I realized we are deep copying the course runs in prepareCourseRuns, and then we technically deep copy them again when we deep copy the entire courseData in prepareCourse so why not just deep copy it once in the beginning and then when prepareCourseRuns needs course_runs, it can have it from the object that has already been copied.

src/components/EditCoursePage/index.jsx
```diff
@@ -126,19 +125,31 @@ class EditCoursePage extends React.Component {
...
+    if (targetRun) {
```
Dillon-Dumesnil: Only deepcopying once will also allow this logic to move back into the `prepareCourseRuns` method, which I think I prefer because it's purpose is deciding which courseRun is the targetRun and that should fall under that helper function.

src/components/EditCoursePage/index.jsx
```diff
@@ -89,32 +85,35 @@ class EditCoursePage extends React.Component {
...
     }];
-    courseData.image = courseData.imageSrc;
+    updatedCourseEditData.image = updatedCourseEditData.imageSrc;
     // We are removing imageSrc so we don't send the image data twice
-    delete courseData.imageSrc;
-    courseData.video = { src: courseData.videoSrc };
-    courseData.key = key;
-    courseData.uuid = uuid;
-    /* eslint-enable no-param-reassign */
+    delete updatedCourseEditData.imageSrc;
```
Dillon-Dumesnil: If we _really_ care about minimal data, you can also delete 
- `updatedCourseEditData.subjectPrimary`
- `updatedCourseEditData.subjectSecondary`
- `updatedCourseEditData.subjectTertiary`
- `updatedCourseEditData.videoSrc`
- `updatedCourseEditData.price`
- `updatedCourseEditData.mode`

src/components/EditCoursePage/index.jsx
```diff
@@ -57,29 +57,39 @@ class EditCoursePage extends React.Component {
...
-    /* eslint-disable no-param-reassign */
-    courseRuns.forEach((courseRun) => {
-      if (courseRun.staff) {
-        courseRun.staff = courseRun.staff.map(staffer => staffer.uuid);
+  prepareSendCourseRunData(courseData, targetRun) {
+    const sendCourseRuns = [];
+    const modifiedCourseRuns =
+      courseData.course_runs.filter(run => !IN_REVIEW_STATUS.includes(run.status));
+
+    // Reduce Staff list to just the UUID
```
Dillon-Dumesnil: This comment should be moved down to above `staff:`

src/components/EditCoursePage/index.jsx
```diff
@@ -57,29 +57,39 @@ class EditCoursePage extends React.Component {
...
   }
 
-  prepareCourseRuns(courseRuns, courseData) {
-    /* eslint-disable no-param-reassign */
-    courseRuns.forEach((courseRun) => {
-      if (courseRun.staff) {
-        courseRun.staff = courseRun.staff.map(staffer => staffer.uuid);
+  prepareSendCourseRunData(courseData, targetRun) {
+    const sendCourseRuns = [];
+    const modifiedCourseRuns =
```
Dillon-Dumesnil: I would add back in the comment about how the backend doesn't like course runs that are in review.

src/components/EditCoursePage/index.jsx
```diff
@@ -57,29 +57,39 @@ class EditCoursePage extends React.Component {
...
+        max_effort: courseRun.max_effort,
+        min_effort: courseRun.min_effort,
+        pacing_type: courseRun.pacing_type,
+        staff: courseRun.staff ? courseRun.staff.map(staffer => staffer.uuid) : courseRun.staff,
+        length: courseRun.length,
+        start: courseRun.start,
+        status: courseRun.status,
+        transcript_languages: courseRun.transcript_languages,
+        weeks_to_complete: courseRun.weeks_to_complete,
+        draft,
```
Dillon-Dumesnil: Everything else is alphabetical, I feel like this should be too? Or is this a linting thing?

src/components/EditCoursePage/index.jsx
```diff
@@ -90,31 +100,48 @@ class EditCoursePage extends React.Component {
...
+      // We want to use the same indicator of draft = false for consistency.
+      if (submittedRun.status === PUBLISHED) {
+        updatingPublishedRun = true;
+      }
+    }
+
+    return {
+      additional_information: courseData.additional_information,
+      draft: !updatingPublishedRun,
+      entitlements: entitlements && entitlements[0] && [{
```
Dillon-Dumesnil: I'm actually not sure we need this `entitlements && entitlements[0] &&` check here because we are not creating draft entitlements for Audit seats.

src/components/EditCoursePage/index.jsx
```diff
@@ -90,31 +100,48 @@ class EditCoursePage extends React.Component {
...
+        price: courseData.price,
+        sku: entitlements[0].sku,
+      }],
+      faq: courseData.faq,
+      full_description: courseData.full_description,
+      image: courseData.imageSrc,
+      key,
+      learner_testimonials: courseData.learner_testimonials,
+      level_type: courseData.level_type,
+      // mode: "verified"
```
Dillon-Dumesnil: remove

src/components/EditCoursePage/index.jsx
```diff
@@ -90,31 +100,48 @@ class EditCoursePage extends React.Component {
...
+      outcome: courseData.outcome,
+      // price: "100.00"
```
Dillon-Dumesnil: Remove

PR #166, opened 2019-05-13
Clearing courseInfo errors on load for course and course run creation  pages so errors from the edit page (or wherever) do not persist when you go to creation.

Dillon-Dumesnil:


src/components/EditCoursePage/index.jsx
```diff
@@ -90,9 +90,6 @@ class EditCoursePage extends React.Component {
       },
     } = this.props;
 
-    // We don't need to send the course runs as part of courseData because they are
-    // pulled out and sent to their own endpoint for saving.
-    delete courseData.course_runs;
```
staubina: This should go away in PR #165

src/components/CreateCoursePage/index.jsx
```diff
@@ -11,6 +11,7 @@ class CreateCoursePage extends React.Component {
   constructor(props) {
     super(props);
     this.handleCourseCreate = this.handleCourseCreate.bind(this);
+    if (props.courseInfo.error) this.props.loadCreateNewCoursePage();
```
staubina: Nit: Could we rename this method?

`loadCreateNewCoursePage`

Maybe `initializeCreateCourse` or `clearErrors`

src/data/actions/courseInfo.js
```diff
@@ -93,9 +103,9 @@ function fetchCourseInfo(id) {
   };
 }
 
-function createCourseRun(course, courseRunData) {
+function createCourseRun(course, courseRunData, fromCourseCreatePage=false) {
```
staubina: Nit: I would name this to be `sCourseCreate` or anything else that does not include `Page` or `from`.

src/data/constants/courseInfo.js
```diff
@@ -2,12 +2,14 @@ const REQUEST_COURSE_INFO_FAIL = 'REQUEST_COURSE_INFO_FAIL';
 const REQUEST_COURSE_INFO_SUCCESS = 'REQUEST_COURSE_INFO_SUCCESS';
 const REQUEST_COURSE_INFO = 'REQUEST_COURSE_INFO';
 const CREATE_COURSE = 'CREATE_COURSE';
+const CREATE_COURSE_LOAD = 'CREATE_COURSE_LOAD';
```
staubina: Would initialize be better here than LOAD?

src/components/CreateCoursePage/index.jsx
Dillon-Dumesnil: I'm hesitant to just do clearErrors because it's specifically clearing courseInfo errors. Any thoughts between `initializeCreateCourse` / `initializeCreateCourseRun` or `clearCourseInfoErrors`? @staubina

staubina:
Should we just close this PR now that you have reworked this?

Dillon-Dumesnil:
Not necessarily because I believe there is still value in clearing out any errors whenever you go to a create page.

staubina:
👍

src/components/CreateCoursePage/index.jsx
```diff
@@ -12,6 +12,9 @@ class CreateCoursePage extends React.Component {
   constructor(props) {
     super(props);
     this.handleCourseCreate = this.handleCourseCreate.bind(this);
+    // We do not want to accidentally persist an error from a different page
+    // when a user comes to a create page.
+    if (props.courseInfo.error) props.clearCourseInfoErrors();
```
mikix: If we're on the create page, don't we want to just clear everything, not just the error? We don't have any use for the data in courseInfo. It'll be overridden in happy path on successful create too, anyway.

Dillon-Dumesnil: Uhmm... Sure, I'll update.

src/components/CreateCourseRunPage/index.jsx
```diff
@@ -12,6 +12,9 @@ class CreateCourseRunPage extends React.Component {
   constructor(props) {
     super(props);
     this.handleCourseCreate = this.handleCourseCreate.bind(this);
+    // We do not want to accidentally persist an error from a different page
+    // when a user comes to a create page.
+    if (props.courseInfo.error) props.clearCourseInfoErrors();
```
jlajoie: nit: I know our linter doesn't care, but I hate one-lining conditionals. Use brackets 😡

src/components/CreateCourseRunPage/index.jsx
```diff
@@ -12,6 +12,9 @@ class CreateCourseRunPage extends React.Component {
   constructor(props) {
     super(props);
     this.handleCourseCreate = this.handleCourseCreate.bind(this);
+    // We do not want to accidentally persist an error from a different page
```
staubina: Nit: If we want this comment to be tied to the method, we should place it there rather than copying it multiple times.

src/components/CreateCoursePage/index.jsx
mikix: If you do this, it could be like, `RESET_COURSE_INFO` or something like that - I think we have another reset action lurking around you could ape.

And you could simply do:
```
case RESET_COURSE_INFO:
    return initialState;
```

Dillon-Dumesnil: So it looks like we can't clear out everything because the createCourseRunPage uses the courseInfo data.

mikix: OK, well then forget everything I said (what does it use?)

Dillon-Dumesnil: `const title = courseInfo.data && courseInfo.data.title ? courseInfo.data.title : '';
 const uuid = courseInfo.data && courseInfo.data.uuid ? courseInfo.data.uuid : '';`

PR #168, opened 2019-05-14
Adding course run creation into the course endpoint so they can be wr…

Dillon-Dumesnil:
…apped in an atomic transaction

### DIFF
--- a/src/components/CreateCoursePage/index.jsx
+++ b/src/components/CreateCoursePage/index.jsx
@@ -1,3 +1,4 @@
+import moment from 'moment';
 import React from 'react';
 import PropTypes from 'prop-types';
 import Helmet from 'react-helmet';
@@ -24,12 +25,12 @@ class CreateCoursePage extends React.Component {
       number: options.number,
       mode: options.enrollmentTrack,
       price: options.price,
+      course_run: {
+        start: moment.utc(options.start).toISOString(),
+        end: moment.utc(options.end).toISOString(),
+      },
     };
-    const courseRunData = {
-      start: options.start,
-      end: options.end,
-    };
-    return this.props.createCourse(courseData, courseRunData);
+    return this.props.createCourse(courseData);
   }
 
   render() {
--- a/src/components/CreateCourseRunPage/index.jsx
+++ b/src/components/CreateCourseRunPage/index.jsx
@@ -1,6 +1,8 @@
+import moment from 'moment';
 import React from 'react';
 import PropTypes from 'prop-types';
 import Helmet from 'react-helmet';
+
 import { CreateCourseRunForm } from './CreateCourseRunForm';
 import LoadingSpinner from '../LoadingSpinner';
 import StatusAlert from '../StatusAlert';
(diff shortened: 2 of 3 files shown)


src/components/CreateCoursePage/index.jsx
```diff
@@ -24,12 +24,13 @@ class CreateCoursePage extends React.Component {
       number: options.number,
       mode: options.enrollmentTrack,
       price: options.price,
+      course_run: {
+        course: `${options.org}+${options.number}`,
+        start: (new Date(options.start)).toISOString(),
```
jlajoie: We imported moment, let's just use that if that's still a part of this repo?

src/components/CreateCoursePage/index.jsx
```diff
@@ -24,12 +24,13 @@ class CreateCoursePage extends React.Component {
       number: options.number,
       mode: options.enrollmentTrack,
       price: options.price,
+      course_run: {
+        course: `${options.org}+${options.number}`,
```
staubina: Why are we constructing this here? I highly recommend we move this to the backend.  And avoid constructing keys on the frontend as much as possible.

staubina: We could also use the key that is generated when the course is generated on the backend.

Dillon-Dumesnil: Yeah, it rubbed me the wrong way too. Will update.

src/components/CreateCourseRunPage/index.jsx
```diff
@@ -27,9 +29,21 @@ class CreateCourseRunPage extends React.Component {
...
+          key,
+          uuid,
+        },
+      },
+      createCourseRun,
     } = this.props;
-    return this.props.createCourseRun(courseInfo.data, options);
+
+    const courseRunData = {
+      course: key,
```
staubina: 👍

src/components/CreateCoursePage/index.jsx
```diff
@@ -24,12 +25,12 @@ class CreateCoursePage extends React.Component {
       number: options.number,
       mode: options.enrollmentTrack,
       price: options.price,
+      course_run: {
+        start: moment.utc(options.start).toISOString(),
```
staubina: 👍

src/data/actions/courseInfo.js
```diff
@@ -93,43 +93,41 @@ function fetchCourseInfo(id) {
...
   return (dispatch) => {
     dispatch(createNewCourse(courseData));
+    dispatch(createNewCourseRun(courseData.course_run));
 
     return DiscoveryDataApiService.createCourse(courseData)
       .then((response) => {
         const course = response.data;
         dispatch(createCourseSuccess(course));
-        dispatch(createCourseRun(course, courseRunData));
+        const courseRun = course.course_runs[0];
```
staubina: Do you need a check here to make sure there is a course_run in the list before accessing it through the index?

staubina: The assumption that there will be one there could be dangerous and cause unexpected errors.

src/data/actions/courseInfo.js
```diff
@@ -93,43 +93,41 @@ function fetchCourseInfo(id) {
...
-      start: (new Date(courseRunData.start)).toISOString(),
-      end: (new Date(courseRunData.end)).toISOString(),
-    };
-    return DiscoveryDataApiService.createCourseRun(data)
+    return DiscoveryDataApiService.createCourseRun(courseRunData)
       .then((response) => {
         const courseRun = response.data;
         dispatch(createCourseRunSuccess(courseRun));
-        return dispatch(push(`/courses/${course.uuid}`));
+        return dispatch(push(`/courses/${courseUuid}`));
```
staubina: Nit: `courseUUID` not a blocker

src/data/actions/courseInfo.js
Dillon-Dumesnil: If the response is successful, I believe it is safe to assume there is a course run since the entire request should fail if there isn't one.

src/data/actions/courseInfo.js
Dillon-Dumesnil: @jlajoie  preferred courseUuid

PR #174, opened 2019-05-21
Ensuring the form goes back to being pristine on save or submit

Dillon-Dumesnil:


src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -145,7 +146,7 @@ class BaseCreateCourseForm extends React.Component {
             type="date"
             component={RenderInputTextField}
             format={value => getDateString(value)}
-            normalize={value => moment.utc(value).toISOString()}
+            normalize={value => moment.utc(value).format(DATE_FORMAT)}
```
mikix: Our API doesn't use ISO format? :(

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -33,13 +33,32 @@ export class BaseEditCourseForm extends React.Component {
     super(props);
 
     this.state = {
-      open: true,
+      open: false,
```
mikix: We starting closed again?

src/data/actions/courseInfo.js
```diff
@@ -68,6 +69,26 @@ function editCourseFail(error) {
...
+    change('imageSrc', initialImageSrc);
+    for (let i = 0; i < initialCourseRuns.length; i += 1) {
+      // eslint-disable-next-line no-param-reassign
+      currentValues.course_runs[i].status = initialCourseRuns[i].status;
+    }
+    // Need to reset courseInfo.courseSaved to false so we don't continuously
+    // overwrite the currentValues.
+    dispatch(clearCourseSaved());
+  };
+}
```
mikix: Will talk in person about all this stuff. I just need some context. Looks gross on the surface, but probably better than all our other options.

src/components/CreateCoursePage/CreateCourseForm.jsx
Dillon-Dumesnil: Not exactly. The standard for ISO is using milliseconds (SS.sss) and we do not use the milliseconds. This allows the fields to match up and pristine to be set. If things change in the future, should be an easy change to go to `toISOString`

src/components/EditCoursePage/EditCourseForm.jsx
Dillon-Dumesnil: Yup. After going through usability testing with partners, we realized they don't seem to intuitively know that the course runs are collapsible components and so we want to try and make the course collapsed by default so they encounter the behavior sooner.

src/data/actions/courseInfo.js
Dillon-Dumesnil: We need to overwrite the current form values because they are changed in the backend so frontend does not have the updated values by default. This will allow the form to return to being pristine after saving.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -33,13 +33,35 @@ export class BaseEditCourseForm extends React.Component {
...
+        course_runs: initialCourseRuns,
+      },
+      updateFrontendAfterSave,
+    } = this.props;
+
+    if (courseSaved) {
+      // We need to overwrite the current form values because they are changed in the
+      // backend so frontend does not have the updated values by default. This will allow
+      // the form to return to being pristine after saving.
+      updateFrontendAfterSave(change, currentValues, initialImageSrc, initialCourseRuns);
```
staubina: Nit: I do not think we need this comment in the code twice.  Within the method is probably enough.

src/data/constants/index.js
```diff
@@ -11,6 +11,8 @@ const UNPUBLISHED = 'unpublished';
 const IN_REVIEW_STATUS = [REVIEW_BY_LEGAL, REVIEW_BY_INTERNAL];
 const POST_REVIEW_STATUSES = [PUBLISHED, REVIEWED];
 
+const DATE_FORMAT = 'YYYY-MM-DD[T]HH:mm:ss[Z]';
```
staubina: 👍

PR #178, opened 2019-05-22
Initial full app enzyme test

staubina:


mikix:
They are looking into cypress and that kind of test for a11y integration testing. But they haven't finished doing that yet.

PR #181, opened 2019-05-24
Fixing failed save when number fields are blank. Also fixes pristine

Dillon-Dumesnil:
saves when going back to empty values.

### DIFF
--- a/src/components/EditCoursePage/index.jsx
+++ b/src/components/EditCoursePage/index.jsx
@@ -80,16 +80,15 @@ class EditCoursePage extends React.Component {
         end: courseRun.end,
         go_live_date: courseRun.go_live_date,
         key: courseRun.key,
-        max_effort: courseRun.max_effort,
-        min_effort: courseRun.min_effort,
+        max_effort: courseRun.max_effort ? courseRun.max_effort : null,
+        min_effort: courseRun.min_effort ? courseRun.min_effort : null,
         pacing_type: courseRun.pacing_type,
         // Reduce Staff list to just the UUID
         staff: courseRun.staff ? courseRun.staff.map(staffer => staffer.uuid) : courseRun.staff,
-        length: courseRun.length,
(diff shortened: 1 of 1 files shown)


src/components/EditCoursePage/index.jsx
```diff
@@ -80,16 +80,15 @@ class EditCoursePage extends React.Component {
         end: courseRun.end,
         go_live_date: courseRun.go_live_date,
         key: courseRun.key,
-        max_effort: courseRun.max_effort,
-        min_effort: courseRun.min_effort,
+        max_effort: courseRun.max_effort ? courseRun.max_effort : null,
+        min_effort: courseRun.min_effort ? courseRun.min_effort : null,
```
mikix: All of these checks will coerce a 0 into a null or ''. That's probably not right?

src/components/EditCoursePage/index.jsx
```diff
@@ -80,16 +80,15 @@ class EditCoursePage extends React.Component {
...
         pacing_type: courseRun.pacing_type,
         // Reduce Staff list to just the UUID
         staff: courseRun.staff ? courseRun.staff.map(staffer => staffer.uuid) : courseRun.staff,
-        length: courseRun.length,
```
mikix: Was that a weeks_to_complete mistake?

Dillon-Dumesnil: Twas.

src/components/EditCoursePage/index.jsx
Dillon-Dumesnil: Turns out we only need to fix this after we receive the data from the backend when it is an int since 0 is false, but '0' is not. Examples of the two checks and the results are below.

0 ? String(0) : '';
> ""

typeof 0 === 'number' ? String(0) : '';
> "0" 

'0' ? '0' : null;
> "0"

PR #182, opened 2019-05-24
Open collapsible after custom error validation

mikix:
Our custom error validation was getting in the way of some fields being able to open the collapsible to focus the field in error.

This fixes that by stopping validation in an error case and having the edit course form notice.

Stopping validation also prevents the submit button from spinning forever.

https://openedx.atlassian.net/browse/DISCO-972
https://openedx.atlassian.net/browse/DISCO-973

I also opened https://openedx.atlassian.net/browse/DISCO-992 to deal with the setTimeout call that I'm lazily introducing.

### DIFF
--- a/src/components/EditCoursePage/EditCourseForm.jsx
+++ b/src/components/EditCoursePage/EditCourseForm.jsx
@@ -40,12 +40,13 @@ export class BaseEditCourseForm extends React.Component {
     this.setCollapsible = this.setCollapsible.bind(this);
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps) {
     const {
       change,
       courseInfo: {
         courseSaved,
       },
+      courseSubmitInfo,
       currentFormValues,
       initialValues: {
         imageSrc: initialImageSrc,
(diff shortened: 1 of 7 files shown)


src/utils/validation.js
```diff
@@ -62,7 +64,9 @@ const getFieldName = (errors) => {
 const handleCourseEditFail = (errors) => {
   const fieldName = getFieldName(errors);
   if (fieldName) {
-    document.getElementsByName(fieldName)[0].focus();
+    setTimeout(() => {
+      document.getElementsByName(fieldName)[0].focus();
+    }, 250);
```
mikix: I hate this, but haven't got a better fix for the react/browser race condition.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -40,12 +40,13 @@ export class BaseEditCourseForm extends React.Component {
     this.setCollapsible = this.setCollapsible.bind(this);
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps) {
```
jlajoie: Didn't know this had args 👍

mikix:
Oh I was just going to land and do any alasdair improvements async later

PR #183, opened 2019-05-24
Adding in image validation for size

Dillon-Dumesnil:


src/components/ImageUpload/index.jsx
```diff
@@ -10,22 +10,39 @@ class ImageUpload extends React.Component {
...
-    this.setState({ value });
-    this.props.input.onChange(value);
+    this.sizeValidator = this.sizeValidator.bind(this);
   }
 
   handleFilePicked(event) {
     const reader = new FileReader();
+    const file = event.target.files[0];
+    // File size is given in bytes
+    const oneMegabyte = 1048576;
```
Dillon-Dumesnil: Thoughts on using the 1024^2 or 1,000,000 definition of Megabyte?

src/components/ImageUpload/index.jsx
```diff
@@ -10,22 +10,39 @@ class ImageUpload extends React.Component {
...
     };
-    reader.readAsDataURL(event.target.files[0]);
+    reader.readAsDataURL(file);
+  }
+
+  sizeValidator(event) {
+    const { onChange } = this.props.input;
+    if (event.target.naturalHeight !== 675 || event.target.naturalWidth !== 1134) {
+      this.setState({ value: '', sizeValidationError: true });
+      onChange('');
```
Dillon-Dumesnil: These two lines are to null out the value of the invalid input so we stop showing the image on the edit form and we won't send it with the form data. The input element will still show the file name of the failed upload.

src/components/ImageUpload/index.jsx
```diff
@@ -46,13 +63,20 @@ class ImageUpload extends React.Component {
       <div className={classNames('form-group', className)}>
         <div name={name} tabIndex={-1}>
           <label htmlFor={id} className="w-100 p-0">{label}</label>  {/* eslint-disable-line jsx-a11y/label-has-for */}
+          {this.state.sizeValidationError &&
+            <StatusAlert
+              alertType="danger"
+              message={'Image size is too large or image dimensions are incorrect. Please upload' +
```
Dillon-Dumesnil: If there are any thoughts on language, let me know. Happy to update.

Dillon-Dumesnil:
If the current image stored is not of the exact dimensions, the error will display without the correct image. Should that be corrected for legacy courses?

src/components/ImageUpload/index.jsx
mikix: As discussed in person, use 1,000,000 definition.

Dillon-Dumesnil: Switched to 1,000,000. Also this is a completely arbitrary number. Let me know if we want to actually figure out what size we allow.

src/components/ImageUpload/index.jsx
```diff
@@ -10,22 +10,43 @@ class ImageUpload extends React.Component {
...
     };
-    reader.readAsDataURL(event.target.files[0]);
+    reader.readAsDataURL(file);
+  }
+
+  sizeValidator(event) {
+    const { pristine } = this.props.meta;
+
+    if (!pristine && (event.target.naturalWidth !== 1134 || event.target.naturalHeight !== 675)) {
+      this.updateValue('', true);
```
mikix: I wish JS had Python-style kwarg support. So you could do `sizeValidationError=true` because unlabeled bool variables are really best when they are the only one like `setPristine(true)` or something. With `updateValue('', true)`, the `true` could be for anything.

One way to avoid this would also fix another UX issue - which is that we could do the user the kindness of telling them which error is the problem. So maybe instead of changing `updateValue`, you add a new method like `setValidationError` or something that takes a string and sets value to '' while also setting a state error string (rather than a state boolean). And then when that string is set, you show the StatusAlert.

Eh?

Dillon-Dumesnil:
So a kind of weird thing now is that we allow saving while the image validation error is being shown and then when you refresh the page, it will show the original image saved. Not sure what desired behaviour would be.
@mikix @staubina

mikix:
On an upload error, can we restore the previous value rather than clearing it out?

Dillon-Dumesnil:
Updated image mocks
image
image

Dillon-Dumesnil:
Round 3:
image
image

image

image

src/components/ImageUpload/index.jsx
```diff
@@ -21,11 +23,52 @@ class ImageUpload extends React.Component {
...
+      },
+    } = this.props;
+
+    if (!pristine && (event.target.naturalWidth !== requiredWidth ||
+        event.target.naturalHeight !== requiredHeight)) {
+      this.updateValue('');
+      this.setState({
+        sizeValidationError: `Image validation failed. Image dimensions are
+        incorrect. Please upload an image with dimensions ${requiredWidth}x${requiredHeight}
+        pixels.`,
```
mikix: How about:
```
That image has the wrong dimensions. Please upload an image with exactly ${requiredWidth}×${requiredHeight} pixels.
```

Dillon-Dumesnil:
Complete with × instead of x @mikix 
image
image

PR #184, opened 2019-05-28
Add OFAC status to Course Runs.

staubina:


src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -295,6 +295,10 @@ class CollapsibleCourseRun extends React.Component {
           disabled={courseInReview}
           required={courseRunSubmitting}
         />
+        <div>
+          <FieldLabel id="number" text="Add OFAC Restriction text to the FAQ section" className="mb-2" />
```
mikix: Might also be worth setting helpText for the FieldLabel so that people that don't know what OFAC is can see a nice tooltip that might calm any confusion.

src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -295,6 +295,10 @@ class CollapsibleCourseRun extends React.Component {
...
+          <div className="mb-3">{courseRun.has_ofac_restrictions ? 'Yes' : 'No'}</div>
```
jlajoie: This is meant to be just a plain text display and not part of the form right? Just verifying.

staubina: Yes, not meant to be editable by partners and we have not added the form actions yet.

PR #190, opened 2019-06-03
Updating Course Run data being sent to backend to include entitlement…

Dillon-Dumesnil:
… mode and price for Seats


src/components/CreateCourseRunPage/index.jsx
```diff
@@ -45,6 +46,8 @@ class CreateCourseRunPage extends React.Component {
       course: key,
       start: options.start,
       end: options.end,
+      mode: entitlements && entitlements[0] && entitlements[0].mode,
+      price: entitlements && entitlements[0] && entitlements[0].price,
```
mikix: Maybe worth a comment like `yes, I know it's weird we only look at first, but the db model allows for multiple, even though we never have multiple` or something?

PR #196, opened 2019-06-11
updates to add program type and name to course run forms

staubina:
[DISCO-944]

changes to support adding expected program type and name

[DISCO-944]: https://openedx.atlassian.net/browse/DISCO-944

### DIFF
--- a/src/components/EditCoursePage/index.jsx
+++ b/src/components/EditCoursePage/index.jsx
@@ -82,6 +82,10 @@ class EditCoursePage extends React.Component {
         content_language: courseRun.content_language,
         draft,
         end: isValidDate(courseRun.end) ? courseRun.end : null,
+        expected_program_type: courseRun.expected_program_type ?
+          courseRun.expected_program_type : null,
+        expected_program_name: courseRun.expected_program_name ?
+          courseRun.expected_program_name : '',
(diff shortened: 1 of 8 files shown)


src/components/EditCoursePage/index.jsx
```diff
@@ -82,6 +82,10 @@ class EditCoursePage extends React.Component {
         content_language: courseRun.content_language,
         draft,
         end: isValidDate(courseRun.end) ? courseRun.end : null,
+        expected_program_type: courseRun.expected_program_type ?
+          courseRun.expected_program_type : null,
+        expected_program_name: courseRun.expected_program_name ?
+          courseRun.expected_program_name : null,
```
mikix: Do you need the tertiary?

src/components/EditCoursePage/index.jsx
```diff
@@ -82,6 +82,10 @@ class EditCoursePage extends React.Component {
         content_language: courseRun.content_language,
         draft,
         end: isValidDate(courseRun.end) ? courseRun.end : null,
+        expected_program_type: courseRun.expected_program_type ?
+          courseRun.expected_program_type : null,
```
jlajoie: is null the intended fallback here instead of an empty string?

PR #198, opened 2019-06-20
ciduarte/DISCO-1010 Adding Time to Start & End Date

ciduarte:
Added new fields to input time alongside dates

### DIFF
--- a/src/components/CreateCoursePage/CreateCourseForm.jsx
+++ b/src/components/CreateCoursePage/CreateCourseForm.jsx
@@ -1,4 +1,3 @@
-import moment from 'moment';
 import React from 'react';
 import { Field, reduxForm } from 'redux-form';
 import PropTypes from 'prop-types';
@@ -11,13 +10,12 @@ import FieldLabel from '../FieldLabel';
 
 import {
   AUDIT_TRACK,
-  DATE_FORMAT,
   ENTITLEMENT_TRACKS,
   PROFESSIONAL_TRACK,
   VERIFIED_TRACK,
 } from '../../data/constants';
 import { endDateHelp, enrollmentHelp, startDateHelp, titleHelp, pacingHelp } from '../../helpText';
-import { getDateString } from '../../utils/index';
+import DateTimeField from '../DateTimeField';
 
 class BaseCreateCourseForm extends React.Component {
   getEnrollmentTrackOptions() {
@@ -147,36 +145,18 @@ class BaseCreateCourseForm extends React.Component {
           <hr />
           <Field
             name="start"
-            type="date"
-            component={RenderInputTextField}
-            format={value => getDateString(value)}
-            normalize={value => moment.utc(value).format(DATE_FORMAT)}
-            label={
-              <FieldLabel
-                id="start-label"
-                text="Start date"
-                required
-                helpText={startDateHelp}
-              />
-            }
-            placeholder="mm/dd/yyyy"
+            component={DateTimeField}
+            dateLabel="Start date"
+            timeLabel="Start time"
+            helpText={startDateHelp}
             required
           />
           <Field
             name="end"
-            type="date"
-            component={RenderInputTextField}
-            format={value => getDateString(value)}
-            normalize={value => moment.utc(value).format(DATE_FORMAT)}
-            label={
-              <FieldLabel
-                id="end-label"
-                text="End date"
-                required
-                helpText={endDateHelp}
-              />
-            }
-            placeholder="mm/dd/yyyy"
+            component={DateTimeField}
+            dateLabel="End date"
+            timeLabel="End time"
+            helpText={endDateHelp}
             required
           />
           <Field
--- a/src/components/CreateCourseRunPage/index.jsx
+++ b/src/components/CreateCourseRunPage/index.jsx
@@ -73,7 +73,6 @@ class CreateCourseRunPage extends React.Component {
       end: options.end,
       pacing_type: options.pacing_type,
     };
-
     return createCourseRun(uuid, courseRunData);
   }
 
--- a/src/components/DateTimeField/index.jsx
+++ b/src/components/DateTimeField/index.jsx
@@ -1,0 +1,120 @@
+import React from 'react';
+import moment from 'moment';
+import PropTypes from 'prop-types';
+import { InputText } from '@edx/paragon';
+import { getDateString, getTimeString } from '../../utils/index';
+import FieldLabel from '../FieldLabel';
+import { DATE_FORMAT } from '../../data/constants';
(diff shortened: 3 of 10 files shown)


src/components/CreateCourseRunPage/index.jsx
```diff
@@ -73,7 +73,6 @@ class CreateCourseRunPage extends React.Component {
       end: options.end,
       pacing_type: options.pacing_type,
     };
-
```
jlajoie: See ya white space 👋

src/components/DateTimeField/index.jsx
```diff
@@ -0,0 +1,119 @@
...
+      date: getDateString(this.props.input.value) || '',
+      time: getTimeString(this.props.input.value) || '',
+    };
+    this.concatDateTime = this.concatDateTime.bind(this);
+    this.updateDate = this.updateDate.bind(this);
+    this.updateTime = this.updateTime.bind(this);
+  }
+
+  concatDateTime(date, time) {
+    const datetime = moment.utc(date + time, 'YYYY/MM/DD HH:mm').format(DATE_FORMAT);
```
jlajoie: nit: Would prefer to do string interpolation here just so it's more clear that this isn't addition and instead we're joining two strings.

src/components/DateTimeField/index.jsx
```diff
@@ -0,0 +1,119 @@
...
+
+    const {
+      date,
+      time,
+    } = this.state;
+
+    return (
+      <div className="row">
+        <div className="col-6">
+          <InputText
```
jlajoie: It looks like we stripped out the 
```
          normalize={value => moment.utc(value).format(DATE_FORMAT)}
```

just making sure this is intentional

src/components/EditCoursePage/index.jsx
```diff
@@ -301,26 +302,27 @@ class EditCoursePage extends React.Component {
...
+        expected_program_type: courseRun.expected_program_type,
+        expected_program_name: courseRun.expected_program_name,
+        go_live_date: courseRun.go_live_date,
+        min_effort: typeof courseRun.min_effort === 'number' ? String(courseRun.min_effort) : '',
+        max_effort: typeof courseRun.max_effort === 'number' ? String(courseRun.max_effort) : '',
+        pacing_type: courseRun.pacing_type,
+        content_language: courseRun.content_language,
+        transcript_languages: courseRun.transcript_languages,
+        weeks_to_complete: typeof courseRun.weeks_to_complete === 'number' ?
+          String(courseRun.weeks_to_complete) : '',
```
jlajoie: Know this isn't your change, but is there a reason we need to cast this to a string? And if it's not a number we just don't render anything? Our DB has this restricted to a smallint already so I'd be really surprised if we passed in something that wasn't a number to this.

src/utils/index.js
```diff
@@ -2,14 +2,18 @@ import moment from 'moment';
...
 
 const getDateString = date => (date ? moment.utc(date).format('YYYY-MM-DD') : '');
+const getTimeString = date => (date ? moment.utc(date).format('HH:mm') : '');
 
 const isValidDate = (dateStr) => {
   const date = moment(dateStr);
   return moment(dateStr) && date.isValid();
 };
 
+const concatDateTime = (dateStr, timeStr) => moment.utc(dateStr + timeStr, 'YYYY/MM/DD HH:mm').format(DATE_FORMAT);
```
jlajoie: Added this comment in above, I think this method would make more sense if we used string interpolation rather than concatenation here as to me it looks like you're adding two dates together.

src/components/DateTimeField/index.jsx
jlajoie: Also looks like you duplicated this method in utils, can delete this here and remove the context binding above too if you import that?

src/components/EditCoursePage/index.jsx
mikix: There was a reason...  @Dillon-Dumesnil would know.

src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -147,37 +145,21 @@ class BaseCreateCourseForm extends React.Component {
...
-                id="start-label"
-                text="Start date"
-                required
-                helpText={startDateHelp}
-              />
-            }
-            placeholder="mm/dd/yyyy"
-            required
+            component={DateTimeField}
+            dateName="startDate"
```
mikix: `dateName` and `timeName` feel like strings that `DateTimeField` could generate itself. Just take the `name` value and append/prepend `date` and `time`.

src/components/DateTimeField/index.jsx
ciduarte: Yeah, this is intentional! The date only needs to get normalized after the date & time strings are concatenated, and because we're using an InputText component rather than a Field, the "value" variable is the equivalent of "format"

src/components/DateTimeField/index.jsx
```diff
@@ -0,0 +1,119 @@
...
+            name={dateName}
+            type="date"
+            value={getDateString(date)}
+            label={
+              <FieldLabel
+                id={`${dateName}-label`}
+                text={dateLabel}
+                required
```
mikix: You dropped the (addmittedly extraneous) `requiredForSubmit` on these labels. No biggy, since they were excess. But just noting.

More importantly, it's not typical for a component to assume that it is required in all forms it's put into. That should probably be a prop.

src/components/DateTimeField/index.jsx
```diff
@@ -0,0 +1,119 @@
...
+                helpText={helpText}
+              />
+            }
+            placeholder="mm/dd/yyyy"
+            required
+            disabled={courseInReview}
```
mikix: You only ever use `courseInReview` to decide `disabled` - you could avoid some of the boundary leakage of a (presumably dumb) date component knowing what a course is if you simply passed in `disabled` as a prop in the first place.

src/components/EditCoursePage/index.jsx
Dillon-Dumesnil: It has to do with the form being set to pristine. It is returned from discovery as an int, but the form treats it as a string so in order for it to match initialValues with currentFormValues, we cast it to a string. The default value is just `''`

src/components/DateTimeField/index.jsx
ciduarte: Do you think it'd be appropriate to remove it from utils and only have this method in the DateTimeField component? I originally had concatDateTime in utils because it was being accessed by three separate files, now it isn't used at all outside of DateTimeField.

PR #199, opened 2019-06-20
Beta Publisher Alert

ciduarte:
Added a banner on publisher home page to alert
users that it is the beta version.

DISCO-1108


src/components/Header/index.jsx
```diff
@@ -20,7 +20,7 @@ class Header extends React.Component {
           <div className="row align-items-center">
             <div className="col-auto justify-content-start">
               <Hyperlink content={this.renderLogo()} destination={process.env.STUDIO_BASE_URL} />
-              <span className="badge badge-secondary beta">Alpha</span>
+              <span className="badge badge-secondary beta">Beta</span>
```
jlajoie: Just double checking this is going live now as part of this PR

PR #201, opened 2019-06-24
Course data will now be sent before Course Run on Edit

staubina:
[DISCO-1087]

[DISCO-1087]: https://openedx.atlassian.net/browse/DISCO-1087

### DIFF
--- a/src/data/services/DiscoveryDataApiService.js
+++ b/src/data/services/DiscoveryDataApiService.js
@@ -75,24 +75,29 @@ class DiscoveryDataApiService {
     return apiClient.options(url);
   }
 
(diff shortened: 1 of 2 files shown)


src/data/services/DiscoveryDataApiService.js
```diff
@@ -75,26 +75,31 @@ class DiscoveryDataApiService {
     return apiClient.options(url);
   }
 
-  static editCourse(courseData, courseRunData) {
-    const { uuid } = courseData;
+  static editCourseRun(courseRunData) {
```
mikix: Maybe for clarity, use a plural here?

PR #204, opened 2019-06-27
Add Course Edit form restrictions

ciduarte:
The following validations were added:
Price ($1 - $10,000)
Start Date: Today or later
End Date: Must be after Start Date
Publish Date: Today or later
Min/Max Hours: greater than zero and less than 168

DISCO-1097

### DIFF
--- a/src/components/CreateCourseRunPage/CreateCourseRunForm.jsx
+++ b/src/components/CreateCourseRunPage/CreateCourseRunForm.jsx
@@ -1,3 +1,4 @@
+import moment from 'moment';
 import React from 'react';
 import { Field, reduxForm } from 'redux-form';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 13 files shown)


src/components/CreateCourseRunPage/CreateCourseRunForm.jsx
```diff
@@ -19,6 +21,7 @@ const BaseCreateCourseRunForm = ({
   uuid,
   getCourseRunOptions,
   parseOptions,
+  currentFormValues,
```
jlajoie: Do we do anything with this?

src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -192,7 +197,11 @@ class CollapsibleCourseRun extends React.Component {
               }
             />
           }
-          extraInput={{ onInvalid: this.openCollapsible }}
+          extraInput={{
+            onInvalid: this.openCollapsible,
+            min: 0,
+            max: 168,
```
jlajoie: Why 168 for max out of curiosity?

ciduarte: Yeah! "Min & Max Effort" ask for hours per week, and that's the max amount of hours in a week.

PR #210, opened 2019-07-02
Fix Start Date bugs and snapshots

ciduarte:
DISCO-1135

### DIFF
--- a/src/components/CreateCoursePage/CreateCourseForm.test.jsx
+++ b/src/components/CreateCoursePage/CreateCourseForm.test.jsx
@@ -25,6 +25,8 @@ const courseRunOptionsReturn = {
 const parseOptionsReturn = [{ label: 'Instructor-paced', value: 'instructor_paced' },
   { label: 'Self-paced', value: 'self_paced' }];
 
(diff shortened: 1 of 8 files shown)


src/components/CreateCoursePage/CreateCourseForm.test.jsx
```diff
@@ -25,13 +25,17 @@ const courseRunOptionsReturn = {
...
+Date.now = jest.fn(() => new Date(Date.UTC(2001, 0, 1)).valueOf());
+
 describe('CreateCourseForm', () => {
   const initialValues = {
     org: 'edx',
     title: 'Hello',
     number: 'edx101',
     enrollmentTrack: 'verified',
     price: 100.00,
+    start: '2001-01-01T12:00:00Z',
```
Dillon-Dumesnil: Are these necessary? They don't seem to show up in the snapshot anywhere outside of the initialValues and currentFormValues so just double checking the reason for adding them. Also commenting because you make the change here, but not in the CreateCourseRunForm.test.jsx file.

ciduarte: Good catch, these were unnecessary.

PR #211, opened 2019-07-02
Don't send status in course run PATCH payload

jmyatt:
in support of DISCO-1080


jmyatt:
closing this as it's not necessary for DISCO-1080 and not necessarily the right thing to do

PR #214, opened 2019-07-03
Optional field labels

ciduarte:
Removed both "required" and "required for submit"
field labels. Added in "optional" labels.

DISCO-1112

### DIFF
--- a/src/components/EditCoursePage/EditCourseForm.jsx
+++ b/src/components/EditCoursePage/EditCourseForm.jsx
@@ -223,6 +223,9 @@ export class BaseEditCourseForm extends React.Component {
             isOpen={open}
             onToggle={this.setCollapsible}
           >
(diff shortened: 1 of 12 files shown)


src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -223,6 +223,9 @@ export class BaseEditCourseForm extends React.Component {
             isOpen={open}
             onToggle={this.setCollapsible}
           >
+            <div className="mb-3">
+              <span className="text-info" aria-hidden> All fields are required for publication unless otherwise specified.</span>
+            </div>
```
mikix: Let's try it with `text-info` but I wonder if using the same color for the "required" text as the "optional" text will be confusing? Probably not, folks will understand.

PR #215, opened 2019-07-03
Reorder fields

ciduarte:
The input fields on edit course page and within collapsible
course runs are being ordered more intuitively

DISCO-1110


ciduarte:
Here are some examples of changes made under course and course runs.

Screen Shot 2019-07-03 at 3 01 31 PM
Screen Shot 2019-07-03 at 3 01 55 PM
Screen Shot 2019-07-03 at 3 02 02 PM

PR #217, opened 2019-07-08
Update tooltips and other helpers

ciduarte:
DISCO-1109

### DIFF
--- a/src/components/EditCoursePage/EditCourseForm.jsx
+++ b/src/components/EditCoursePage/EditCourseForm.jsx
@@ -253,7 +253,7 @@ export class BaseEditCourseForm extends React.Component {
                   id="mode.label"
                   text="Enrollment track"
                   helpText={enrollmentHelp}
-                  extraText="(Cannot edit after submission)"
+                  extraText="Cannot edit after submission"
                 />
               }
               extraInput={{ onInvalid: this.openCollapsible }}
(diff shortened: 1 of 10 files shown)


ciduarte:
For reference, "extra text" was reformatted to appear under the label like this:

Screen Shot 2019-07-08 at 1 43 54 PM

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -307,6 +308,7 @@ export class BaseEditCourseForm extends React.Component {
                       </p>
                     </div>
                   }
+                  extraText="Image must be 1134x675 pixels in size."
```
mikix: This does feel like we can pull this out of the tooltip now? Or did Carol want it in both places?

src/components/FieldLabel/index.jsx
```diff
@@ -21,7 +21,7 @@ const FieldLabel = ({
       <strong>{text}</strong>
       {helpText && id && <FieldHelp id={`${id}-help`} tip={helpText} />}
       {requireText}
-      {extraText}
+      <div><span className="text-muted">{extraText}</span></div>
```
mikix: Eh, you *could* put the className on the div, not the span. But it's all the same.

src/components/EditCoursePage/EditCourseForm.jsx
ciduarte: Good question! Checked with Carol and she said both places

PR #218, opened 2019-07-08
DISCO-1152 use change function for course status like we do for image…

xnick421:
…Src when updating


src/data/actions/courseInfo.js
```diff
@@ -91,10 +91,7 @@ function updateFormValuesAfterSave(change, currentFormValues, initialImageSrc, i
     // This emits a redux action called CHANGE that will update currentFormValues.imageSrc
     change('imageSrc', initialImageSrc);
     for (let i = 0; i < initialCourseRuns.length; i += 1) {
-      // Status is modified directly because it is not a registeredField with the form, but
-      // is still used in comparison for deciding if the form is pristine.
-      // eslint-disable-next-line no-param-reassign
-      currentFormValues.course_runs[i].status = initialCourseRuns[i].status;
+      change(`course_runs[${i}].status`, initialCourseRuns[i].status);
```
Dillon-Dumesnil: Interesting. I didn't think that would work because it isn't a Field and I thought you could only use `change` on Field values. Nice job!

PR #219, opened 2019-07-09
Confirmation Modal, Small UI Fixes

ciduarte:
ConfirmationModal (previously SubmitConfirmModal) set to
appear on both Course Creation and Course Run Submission.
UI fixes include links opening in a new tab and changes
to field labels and Staffer Form status alert

DISCO-1111


src/components/StafferPage/index.jsx
```diff
@@ -155,19 +155,19 @@ class StafferPage extends React.Component {
...
         </Helmet>
-
+        { referrer &&
+        <StatusAlert
+          id="sent-from-edit-course-info"
+          alertType="info"
+          message="The data you entered on the course edit screen is saved. You will return to that page when you have finished updating instructor information."
+          dismissible
+        />
+        }
```
mikix: Why pull this out? When ouside the `PageContainer`, it stretches full page width and the text gets very long before wrapping.

I know we have them outside the PageContainer elsewhere, but I personally consider those instances mistakes.

ciduarte: Good catch! For some reason I assumed it was the other way around, but it would probably look better if they were all within the PageContainer.

PR #220, opened 2019-07-09
Support viewing courses we can't edit

mikix:
* Ask for OPTIONS on the list-courses endpoint, not a specific course endpoint - because the backend will refuse an OPTIONS for a specific course if you can't PUT to it.
* Support the new `editable` serialized field and disable all fields / hide buttons when it is `false` 

https://openedx.atlassian.net/browse/DISCO-1143

### DIFF
--- a/src/components/EditCoursePage/CollapsibleCourseRun.jsx
+++ b/src/components/EditCoursePage/CollapsibleCourseRun.jsx
@@ -109,6 +109,7 @@ class CollapsibleCourseRun extends React.Component {
       courseInReview,
       courseRun,
       courseSubmitting,
+      editable,
       isSubmittingForReview,
       languageOptions,
       programOptions,
@@ -128,6 +129,8 @@ class CollapsibleCourseRun extends React.Component {
     // Checks if the current course run is the one triggering submission for review
     const courseRunSubmitting = !courseRunInReview && isSubmittingForReview && targetRun &&
       (targetRun.key === courseRun.key);
+
+    const disabled = courseInReview || !editable;
 
     return (
       <Collapsible
@@ -169,7 +172,7 @@ class CollapsibleCourseRun extends React.Component {
               getDateString(courseRun.go_live_date) : getDateString(moment()),
           }}
           placeholder="mm/dd/yyyy"
-          disabled={courseInReview}
+          disabled={disabled}
           required={courseRunSubmitting}
         />
         <Field
@@ -178,7 +181,7 @@ class CollapsibleCourseRun extends React.Component {
           dateLabel="Start date"
           timeLabel="Start time"
           helpText={startDateHelp}
-          disabled={courseInReview}
+          disabled={disabled}
           required
           minDate={moment(courseRun.start).isBefore(moment()) ?
             getDateString(courseRun.start) : getDateString(moment())}
@@ -190,7 +193,7 @@ class CollapsibleCourseRun extends React.Component {
           dateLabel="End date"
           timeLabel="End time"
           helpText={endDateHelp}
-          disabled={courseInReview}
+          disabled={disabled}
           required
           minDate={getDateString(moment(courseRun.start).add(1, 'd') || moment())}
           onInvalid={this.openCollapsible}
@@ -209,7 +212,7 @@ class CollapsibleCourseRun extends React.Component {
             />
           }
           extraInput={{ onInvalid: this.openCollapsible }}
-          disabled={courseInReview}
+          disabled={disabled}
           required={courseRunSubmitting}
         />
         <FieldLabel
@@ -230,7 +233,7 @@ class CollapsibleCourseRun extends React.Component {
           name={`${courseId}.staff`}
           component={StaffList}
           extraInput={{ onInvalid: this.openCollapsible }}
-          disabled={courseInReview}
+          disabled={disabled}
           courseRunKey={courseRun.key}
           owners={owners}
           courseUuid={courseUuid}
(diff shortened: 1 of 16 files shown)


src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -450,6 +456,7 @@ CollapsibleCourseRun.defaultProps = {
   courseSubmitInfo: {
     isSubmittingRunReview: false,
   },
+  editable: true,
```
Dillon-Dumesnil: Just want to call out and make sure having it default to `true` is intentional. If for whatever reason, our data is messed up, I think I'm of the opinion we should lean towards locking the form.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -99,15 +99,15 @@ export class BaseEditCourseForm extends React.Component {
       return [];
     }
 
-    return data.actions.PUT;
+    return data.actions.POST;
```
Dillon-Dumesnil: ?

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -786,25 +789,27 @@ export class BaseEditCourseForm extends React.Component {
             courseSubmitting={submitting}
             {...this.props}
           />
-          {this.getAddCourseRunButton(courseInReview, pristine, uuid)}
+          {this.getAddCourseRunButton(disabled, pristine, uuid)}
+          {editable &&
           <ButtonToolbar className="mt-3">
```
Dillon-Dumesnil: Nit: Indent this after add in the `{editable &&` Unless that goes against eslint. I can't ever remember.

src/components/EditCoursePage/index.jsx
```diff
@@ -382,6 +383,11 @@ class EditCoursePage extends React.Component {
         />
 
         <PageContainer>
+          { showForm && !editable && <StatusAlert
+            alertType="secondary"
+            message="You have permission to view this course, but not edit. If you would like to edit the course, please contact your project coordinator."
```
Dillon-Dumesnil: This might only be a temporary message, but we should make sure we remember that eventually they should be able to just contact a current editor on the course to be added as one.

src/components/EditCoursePage/index.jsx
```diff
@@ -433,6 +439,7 @@ class EditCoursePage extends React.Component {
                 owners={owners}
                 isSubmittingForReview={isSubmittingForReview}
                 targetRun={targetRun}
+                editable={editable}
```
Dillon-Dumesnil: Thoughts on passing in a `disabled` prop that you define here rather than have to redefine `disabled = courseInReview | ~editable` in the edit form and collapsible course runs.

src/components/ImageUpload/index.jsx
```diff
@@ -132,6 +140,7 @@ ImageUpload.propTypes = {
 
 ImageUpload.defaultProps = {
   className: '',
+  disabled: false,
```
Dillon-Dumesnil: Just a reminder to update this if you end up updating the default value in the form.

src/containers/EditCourse/index.jsx
```diff
@@ -33,7 +33,7 @@ const mapDispatchToProps = {
 const mergeProps = (stateProps, actionProps, { id }) => ({
   ...stateProps,
   fetchCourseInfo: () => actionProps.fetchCourseInfo(id),
-  fetchCourseOptions: () => actionProps.fetchCourseOptions(id),
+  fetchCourseOptions: () => actionProps.fetchCourseOptions(),
```
Dillon-Dumesnil: ?

src/components/EditCoursePage/EditCourseForm.jsx
mikix: This is because OPTIONS returns the set of actions for the default expected HTTP verb. On a course-detail endpoint, that's PUT (update course), but on a course-list endpoint, that's POST (new course).

src/components/EditCoursePage/EditCourseForm.jsx
mikix: I think eslint accepts both. I figured this was easier to visually follow, but I grant it makes it harder to see what is controlled by the editable.

mikix: (Meaning, I'll make the indent change)

src/components/EditCoursePage/index.jsx
mikix: Yes. I figured when we add the editor UI, we can fix this up.

src/components/EditCoursePage/index.jsx
mikix: I did it this way because the CollapsibleCourseRun wants to know editable status standalone (so it knows whether to show the submit button). So it is interested in both fields separately. I guess I could pass in `disabled` *plus* `editable`, so it doesn't have to calculate it... But then it feels weird that it ignores the `disabled` flag for the submit button sometimes. If it decides what is disabled, it feels less odd that it ignores orders for that one button.

src/containers/EditCourse/index.jsx
mikix: I dropped the id argument to the `fetchCourseOptions` action, because now we just hit the list endpoint.

src/components/ImageUpload/index.jsx
mikix: I still think this is a fine default. This is going to be overturned by parent component all the time (which now defaults to disabled=True). As a component-level default, I think being enabled by default makes sense.

PR #221, opened 2019-07-09
DISCO-656: Adds in search to dashboard

jlajoie:
Adds in search bar to dashboard, depends on course-discovery API accepting `?pubq=`

Without search applied:
Screen Shot 2019-07-09 at 1 12 43 PM

With search applied:
Screen Shot 2019-07-09 at 1 12 50 PM


src/data/actions/table.js
```diff
@@ -88,50 +116,14 @@ const paginateTable = (tableId, fetchMethod, pageNumber) => (
   }
 );
 
-const sortBy = (dataToSort, orderField) => {
```
jlajoie: Wasn't used because we never set num_pages

src/data/actions/table.js
```diff
@@ -88,50 +116,14 @@ const paginateTable = (tableId, fetchMethod, pageNumber) => (
...
-    const tableState = getState().table[tableId];
+  (dispatch) => {
     const options = {
       ...getPageOptionsFromUrl(),
       ordering,
     };
     dispatch(sortRequest(tableId, ordering));
 
-    // If we can sort client-side because we have all of the data, do that
-    if (tableState.data && tableState.data.num_pages === 1) {
```
jlajoie: Wasn't used because we never set num_pages

src/components/ButtonToolbar/index.jsx
```diff
@@ -5,7 +5,7 @@ import classNames from 'classnames';
 const ButtonToolbar = (props) => {
   const { className } = props;
   const fullClassName = classNames(
-    'btn-toolbar justify-content-end',
+    'btn-toolbar',
```
mikix: Dropping that might change all the other toolbars on the site. We use it on the course edit page and the course run collapsibles, I think.

jlajoie: Oh good catch, I should have seen if this had cascading changes, I'll look for a better way to override

PR #223, opened 2019-07-15
Disallow adding new runs when in review

mikix:
If a course is under review, don't allow the workaround of manually editing the URL field and putting in /rerun to still get the form.

Now you'll see a little warning page instead explaining that you can't add new runs.

https://openedx.atlassian.net/browse/DISCO-894

Screenshot from 2019-07-15 12-19-43

### DIFF
--- a/src/components/CreateCourseRunPage/CreateCourseRunPage.test.jsx
+++ b/src/components/CreateCourseRunPage/CreateCourseRunPage.test.jsx
@@ -1,7 +1,9 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 
+import { CreateCourseRunForm } from './CreateCourseRunForm';
 import CreateCourseRunPage from './index';
+import StatusAlert from '../StatusAlert';
 
 describe('CreateCourseRunPage', () => {
   it('renders html correctly', () => {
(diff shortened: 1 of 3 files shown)


src/components/CreateCourseRunPage/CreateCourseRunPage.test.jsx
```diff
@@ -52,4 +54,30 @@ describe('CreateCourseRunPage', () => {
     />);
     expect(component).toMatchSnapshot();
   });
+
+  it('refuses when course is under review', () => {
```
jlajoie: tiny nit: refuses what

src/components/CreateCourseRunPage/index.jsx
```diff
@@ -1,7 +1,9 @@
+/* eslint-disable camelcase */
```
jlajoie: What in this file is causing us to need this disabled out of curiosity?

src/components/CreateCourseRunPage/CreateCourseRunPage.test.jsx
mikix: Fair, will clarify

src/components/CreateCourseRunPage/index.jsx
mikix: Pulling out `course_runs` from the serialized data. Then `course_runs` is a JS variable, and eslint doesn't like that. We do same thing in main course form too.

PR #226, opened 2019-07-25
Ndalfonso/disco 1205 curtain launch tweaks

xnick421:


src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -18,9 +18,18 @@ import {
 } from '../../data/constants';
 import { endDateHelp, enrollmentHelp, startDateHelp, titleHelp, pacingHelp } from '../../helpText';
 import DateTimeField from '../DateTimeField';
-import { getDateString } from '../../utils';
+import { getDateString, localTimeZone } from '../../utils';
 
 class BaseCreateCourseForm extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { organizations: [{ label: 'Select organization', value: '' }] };
```
ciduarte: Do you need to use the state to handle the organization list? I think the original `processOrganizations()` method handled it as expected

src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -62,7 +70,7 @@ class BaseCreateCourseForm extends React.Component {
           <Field
             name="org"
             component={RenderSelectField}
-            options={this.processOrganizations(organizations)}
+            options={this.state.organizations}
```
ciduarte: (Similar to above) Not sure if this change is necessary in order to default to the only organization in the list

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -151,9 +151,9 @@ export class BaseEditCourseForm extends React.Component {
     this.setCollapsible(true);
   }
 
-  parseOptions(inChoices) {
-    return inChoices.map(choice =>
-      ({ label: choice.display_name, value: choice.value }));
+  parseOptions(choices, isLanguages) {
+    const options = choices.map(choice => ({ label: choice.display_name, value: choice.value }));
+    return isLanguages ? options.sort((a, b) => (a.label.localeCompare(b.label))) : options;
```
ciduarte: I think this can be done more easily in the backend, sorting them when we retrieve the languages.

We should also avoid overloading this method with two different purposes. If we're going to keep the sorting in the frontend, we should consider creating a `parseLanguageOptions()` to handle it separately.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -792,6 +802,7 @@ export class BaseEditCourseForm extends React.Component {
           {this.getAddCourseRunButton(disabled, pristine, uuid)}
           {editable &&
             <ButtonToolbar className="mt-3">
+              {submitState === 'default' ? cancelButton : null}
```
ciduarte: I would double check with Carol, but it feels like `cancelButton` should always be available, unless the form is submitting (Similar to how the save button works `disabled={disabled || submitting}`)

This would allow the user to click 'Cancel' and return to the CourseTable page, regardless of whether they've made a change in the form.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -55,6 +58,11 @@ export class BaseEditCourseForm extends React.Component {
       updateFormValuesAfterSave,
     } = this.props;
 
+    if (initialCourseRuns.length && !this.state.collapsiblesOpen.length) {
+      const arr = initialCourseRuns.map(() => (false));
```
Dillon-Dumesnil: Nit: name this variable something more helpful. Maybe just calling it `const collapsiblesOpen =`

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -72,6 +80,17 @@ export class BaseEditCourseForm extends React.Component {
     }
   }
 
+  resetOpenCourseRuns() {
```
Dillon-Dumesnil: Nit: rename to `closeOpenCourseRuns`. Reset makes it sound like resetting data or something and this doesn't do that.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -216,6 +234,19 @@ export class BaseEditCourseForm extends React.Component {
...
+    const cancelButton = (
+        <button
+          onClick={() => {
+            this.setCollapsible(false);
+            this.props.reset();
+            this.resetOpenCourseRuns();
+          }}
+          className="btn btn-outline-primary"
+        >
+          Cancel
```
Dillon-Dumesnil: I think `Cancel` on its own is misleading in this context. We are now clearing data instead of leaving the form. I would argue for something like `Cancel Edits` or `Clear Edits`

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -216,6 +234,19 @@ export class BaseEditCourseForm extends React.Component {
 
     const disabled = courseInReview || !editable;
 
+    const cancelButton = (
+        <button
+          onClick={() => {
+            this.setCollapsible(false);
+            this.props.reset();
```
Dillon-Dumesnil: Instead of calling `this.props.reset()`, you should pull `reset` out when we are deconstructing props.

PR #227, opened 2019-07-25
DISCO-1201: FLEX for instructor position org

Dillon-Dumesnil:
Also removes stafferOptions as they are no longer used


src/components/StafferPage/StafferForm.jsx
```diff
@@ -107,10 +93,11 @@ const BaseStafferForm = ({
...
         />
         <Field
-          name="position.organization_id"
-          component={RenderSelectField}
-          options={extractOrgChoices(stafferOptions)}
+          name="position.organization_override"
+          component={RenderInputTextField}
+          type="text"
           label={<FieldLabel text="Organization" />}
+          extraInput={{value: position.organization_override || position.organization_name}}
```
mikix: Interesting - I might have expected the serializer to abstract this detail away for us, but alright.

Which is a better philosophical position? That the serializer should hide these details or expose them?

Dillon-Dumesnil: You were write. I thought organization_name was coming from the Foreign key, but it does take into account org_override as well.

Dillon-Dumesnil:
Is it weird to have initialValues in the defaultPropValues and PropTypes?

src/components/StafferPage/StafferForm.jsx
```diff
@@ -107,10 +92,11 @@ const BaseStafferForm = ({
...
         />
         <Field
-          name="position.organization_id"
-          component={RenderSelectField}
-          options={extractOrgChoices(stafferOptions)}
+          name="position.organization_override"
+          component={RenderInputTextField}
+          type="text"
           label={<FieldLabel text="Organization" />}
+          extraInput={{ value: organizationName }}
```
ciduarte: Out of curiosity - are you using extraInput to set the default value of this field? If so, is this sufficient to register that value within the form? I would've assumed we'd want to pass `organizationName` through the initialValues.

I think I'm confused on the difference between 'position.organization_override' and 'data.position.organization_name'..

Dillon-Dumesnil: So it’s just to set the default display value. The form value is only going to be based on data.position.organization_override, but in case they already have an organization defined, we want to just use the name from that org instead of showing it blank since the override is blank. Then if/when they update the value, it will be put into the Form as `position.organization_override`

PR #228, opened 2019-07-26
Allow user to select run to re-run

ciduarte:
DISCO-1219


src/components/DateTimeField/index.jsx
```diff
@@ -11,7 +11,7 @@ class DateTimeField extends React.Component {
     super(props);
     this.state = {
       date: getDateString(this.props.input.value) || '',
-      time: getTimeString(this.props.input.value) || '',
+      time: getTimeString(this.props.input.value) || '12:00',
```
ciduarte: Quick fix while I was in there, sets default form value to 12:00PM. Previous commit that addressed this only set the field's display text to 12:00PM.

src/components/CreateCourseRunPage/index.jsx
```diff
@@ -83,6 +86,26 @@ class CreateCourseRunPage extends React.Component {
...
+    } else {
+      c = 0;
+    }
+    return c;
+  }
+
+  parseCourseRunLabels(courseRuns) {
+    const sortedCourseRuns = courseRuns.sort(this.compareDates);
+    return sortedCourseRuns.map((run) => {
+      const runKey = (run.key).split('+', 3)[2];
```
ciduarte: This is operating on the assumption that our course key will always be in this format: `course-v1:edX+cs101+2T2019`
Is this a safe assumption, or is there another way I can grab the course run's ID?

src/components/CreateCoursePage/index.jsx
```diff
@@ -64,6 +64,7 @@ class CreateCoursePage extends React.Component {
         start: options.start,
         end: options.end,
         pacing_type: options.pacing_type,
+        rerun: '',
```
Dillon-Dumesnil: You should be able to get away with not supplying this at all in the course creation page by just having the backend do a `request.data.get('rerun', '')` (which will default to `''`)

src/components/CreateCourseRunPage/CreateCourseRunForm.jsx
```diff
@@ -93,6 +115,10 @@ const BaseCreateCourseRunForm = ({
 
 BaseCreateCourseRunForm.defaultProps = {
   currentFormValues: {},
+  courseRunLabels: [],
+  initialValues: {
```
Dillon-Dumesnil: Why is initialValues in the defaultProps and propTypes if it isn't being used as a prop?

Ahh. I see, you're passing it in. But it's kinda handled behind the scenes. I'd remove it out of here. We don't specify initialValues in the props anywhere else.

src/components/CreateCourseRunPage/index.jsx
```diff
@@ -83,6 +86,18 @@ class CreateCourseRunPage extends React.Component {
       ({ label: choice.display_name, value: choice.value }));
   }
 
+  compareDates(a, b) {
```
Dillon-Dumesnil: Don't love the a, b. Maybe even `dateOne` and `dateTwo` or something. Also, looking at what this is doing, it is misleading to say it is comparing dates because it isn't just comparing two dates. It is specifically comparing two start dates on an object that includes a key called start and that field is a date.

Dillon-Dumesnil: I think this should also have a docstring to explain what this is going to return. It's unclear from the function name and the code inside the function. This might also be a good opportunity to use an arrow function definition. I think it would look something like 
```
const compareDates = (a, b) => (moment(b.start).diff(moment(a.start)))
```

src/components/CreateCourseRunPage/index.jsx
```diff
@@ -83,6 +86,18 @@ class CreateCourseRunPage extends React.Component {
...
+    return (moment(b.start).diff(moment(a.start)));
+  }
+
+  parseCourseRunLabels(courseRuns) {
+    const sortedCourseRuns = courseRuns.sort(this.compareDates);
```
Dillon-Dumesnil: That's pretty cool. I haven't seen this before. Nice!

src/components/CreateCourseRunPage/index.jsx
Dillon-Dumesnil: It is not unfortunately. Old course keys used to be `Org/Number/Term`. I think you might need to write a regex or something to parse this. I'm not sure off the top of my head. I would maybe poke around discovery since I know we do some parsing there and it should have a way of doing it. Maybe in the course runs endpoint.

Dillon-Dumesnil: I would also rename this variable to be `runTerm`. The thing that you are parsing is already the runKey.

src/components/CreateCourseRunPage/index.jsx
```diff
@@ -116,6 +131,8 @@ class CreateCourseRunPage extends React.Component {
     const showSpinner = !startedFetching || courseInfo.isFetching || courseRunOptions.isFetching;
     const showForm = startedFetching && !courseInfo.isFetching && !courseRunOptions.isFetching &&
       !courseInReview;
+    const sortedRunLabels = course_runs && this.parseCourseRunLabels(course_runs);
+    const defaultRun = sortedRunLabels && sortedRunLabels[0].value;
```
Dillon-Dumesnil: Since this won't have a default value in the form, you could update this to
`const defaultRun = (sortedRunLabels && sortedRunLabels[0].value) || '';`

src/components/DateTimeField/index.jsx
Dillon-Dumesnil: Good catch. Nice job!

src/components/EditCoursePage/index.jsx
```diff
@@ -92,6 +92,7 @@ class EditCoursePage extends React.Component {
         min_effort: courseRun.min_effort ? courseRun.min_effort : null,
         pacing_type: courseRun.pacing_type,
         // Reduce Staff list to just the UUID
+        rerun: courseRun.rerun ? courseRun.rerun : null,
```
Dillon-Dumesnil: Move this so the comment still applies to the line directly below it.

src/components/CreateCourseRunPage/index.jsx
ciduarte: Do you think it'd be more appropriate to do something like this?

```
parseCourseRunLabels(courseRuns) {
    // Sort course runs by descending start dates
    const sortedCourseRuns =
      courseRuns.sort((run1, run2) => moment(run2.start).diff(moment(run1.start)));
```

Or maybe this:
```
// Helper method used to compare course runs and sort by descending start dates
compareRunDates = (run1, run2) => (moment(run2.start).diff(moment(run1.start)));

parseCourseRunLabels(courseRuns) {
    const sortedCourseRuns = courseRuns.sort(this.compareRunDates);
```

I can't tell what the tradeoffs are between these two.

ciduarte: And do you think I need to describe what that `moment().diff();` is returning? I'm not sure if the Array.sort() parameters are common knowledge..

src/components/EditCoursePage/index.jsx
ciduarte: Good catch

src/components/CreateCourseRunPage/index.jsx
Dillon-Dumesnil: I think I prefer the top one, unless you think this compareRunDates function could be used in other parts of PubFE, in which case I'd argue for putting it in a util file. If you do that, I'd just write comments in both places.

Also, if you keep compareRunDates, please rename to compareRunStartDates since it doesn't compare any other dates. Or abstract it so it can compare any two dates (and if you do this, maybe also add in a parameter to specify ascending vs descending order). I don't have a preference either way.

Re needing to describe: Eh, I think your comment does enough. If someone needs to know more, they can look up the docs. You describe the functionality of your code by saying that you're sorting course runs by descending start dates. Would be overkill to copy moment.js docs or Array.prototype docs.

src/components/CreateCourseRunPage/index.jsx
```diff
@@ -83,6 +86,16 @@ class CreateCourseRunPage extends React.Component {
       ({ label: choice.display_name, value: choice.value }));
   }
 
+  parseCourseRunLabels(courseRuns) {
+    // Sort course runs by descending start dates
+    const sortedCourseRuns =
+      courseRuns.sort((run1, run2) => moment(run2.start).diff(moment(run1.start)));
+    return sortedCourseRuns.map((run) => {
+      const runTerm = (run.key).split(/\+|\//).pop();
```
Dillon-Dumesnil: Nit: you shouldn't need the wrapping parens on the run.key

ciduarte: Good catch!

PR #230, opened 2019-07-31
DISCO-1222: Allow changing Course Number post creation

Dillon-Dumesnil:
Also fixed our `make lint` command so both subcommands run inside of the container instead of only the first.


src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -281,10 +280,46 @@ export class BaseEditCourseForm extends React.Component {
...
-            </div>
+            <Field
+              name="number"
+              component={RenderInputTextField}
+              type="text"
+              label={
+                <FieldLabel
+                  id="number.label"
+                  text="Number"
+                  helpText={
```
ciduarte: Might be worth putting this in the helpText file, since this tool tip is also used in the CreateCourseForm.

Dillon-Dumesnil: Nice call! I had no idea that file existed. Thanks Carla!

PR #232, opened 2019-08-05
Updates to side pane UI

ciduarte:
Added a cancel button to 'add/set editor',
made the select field a paragon component,
CSS edits made to clean up trickle down of
PageContainer on other pages.

DISCO-1050

### DIFF
--- a/src/components/CreateCourseRunPage/index.jsx
+++ b/src/components/CreateCourseRunPage/index.jsx
@@ -129,8 +129,9 @@ class CreateCourseRunPage extends React.Component {
     const showSpinner = !startedFetching || courseInfo.isFetching || courseRunOptions.isFetching;
     const showForm = startedFetching && !courseInfo.isFetching && !courseRunOptions.isFetching &&
       !courseInReview;
-    const sortedRunLabels = course_runs && this.parseCourseRunLabels(course_runs);
-    const defaultRun = sortedRunLabels ? sortedRunLabels[0].value : '';
+    const sortedRunLabels = (showForm &&
+      (course_runs ? this.parseCourseRunLabels(course_runs) : []));
+    const defaultRun = (showForm && (sortedRunLabels.length ? sortedRunLabels[0].value : ''));
 
     return (
       <React.Fragment>
--- a/src/components/SidePanes/UsersPane.jsx
+++ b/src/components/SidePanes/UsersPane.jsx
@@ -1,10 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { Icon } from '@edx/paragon';
+import { Icon, InputSelect } from '@edx/paragon';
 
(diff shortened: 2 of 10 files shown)


ciduarte:
Screen Shot 2019-08-05 at 10 01 50 AM
Screen Shot 2019-08-05 at 10 02 04 AM

src/components/SidePanes/UsersPane.jsx
```diff
@@ -47,6 +49,13 @@ class UsersPane extends React.Component {
     });
   }
 
+  cancelAddUser() {
+    this.setState({
+      addingUser: false,
+      newEditorChoice: 0,
```
Dillon-Dumesnil: What does setting this to `0` mean? Like, I know it is supposed to be a user id, but is there a chance we are going to send this 0 anywhere? If so, what is the impact of that? Is it better to send `undefined` or `null`?

src/components/SidePanes/UsersPane.test.jsx
```diff
@@ -112,9 +112,6 @@ describe('UsersPane', () => {
     const startAddButton = wrapper.find('.usersPane-startAdd');
     startAddButton.simulate('click');
 
-    const select = wrapper.find('select');
```
Dillon-Dumesnil: Is this not possible to keep anymore? I could see this check still having value.

src/components/CreateCourseRunPage/index.jsx
```diff
@@ -129,8 +129,9 @@ class CreateCourseRunPage extends React.Component {
     const showSpinner = !startedFetching || courseInfo.isFetching || courseRunOptions.isFetching;
     const showForm = startedFetching && !courseInfo.isFetching && !courseRunOptions.isFetching &&
       !courseInReview;
-    const sortedRunLabels = course_runs && this.parseCourseRunLabels(course_runs);
-    const defaultRun = sortedRunLabels ? sortedRunLabels[0].value : '';
+    const sortedRunLabels = (showForm &&
```
ciduarte: Found a bug where the select field on the 'add course run' page will cause an error for courses without existing runs. Wanted to push a fix for this edge case out with these edits prior to bug bash.

src/components/SidePanes/UsersPane.test.jsx
ciduarte: I was actually able to select the option rather than the entire select field.

If you're curious, follow up to your suggestion of debugging the wrapper: I was running into issues using `wrapper.find()` because the InputSelect component from Paragon is actually labeled as `<asInput(withDeprecatedProps(InputSelect)) />`, so neither `wrapper.find('select')` or `wrapper.find('InputSelect')` was doing the trick. Selecting the option directly is more straight forward, and will (hopefully) help if/when we switch to Paragon V2.

PR #235, opened 2019-08-06
DISCO-1188 support ofac restriction, ofac comment, and status

xnick421:


src/data/actions/courseInfo.js
```diff
@@ -174,6 +180,32 @@ function createCourse(courseData) {
   };
 }
 
+function handleCourseRuns(dispatch, courseRunData, course, submitReview) {
+  // status property will be present for internal review process because we pass the run
```
ciduarte: nit: Clean up this comment's wording so it clarifies the functionality more quickly.
```We'll use the presence of the course run status property to signal an internal review process because it is only present when we pass in a single run for internal review, rather than an array of runs on a regular edit```

File: src/components/EditCoursePage/CollapsibleCourseRun.jsx [hunk 2]
```diff
@@ -145,6 +149,34 @@ class CollapsibleCourseRun extends React.Component {
...
+      default:
+        return '--';
+    }
+  }
+
+  displayDefaultButtonLabel(status, isAdmin) {
+    if (isAdmin) {
+      switch (status) {
+        case REVIEW_BY_LEGAL:
+          return 'Save & Send to PC Review';
```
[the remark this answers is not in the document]
xnick421: this line is only viewed by staff members, never regular users.  line 177 is for regular users.  i  dont think the language has to match because the the labels are for 2 totally separate flows.  Also, this is what is on the mock.

src/data/actions/courseInfo.js
xnick421: will change now :)

src/data/services/DiscoveryDataApiService.js
```diff
@@ -114,6 +114,21 @@ class DiscoveryDataApiService {
...
   }
 
+  static internalReviewEdit(courseRun) {
+    const queryParams = {
+      exclude_utm: 1,
+    };
+    const { key } = courseRun;
+    const body = Object.assign({}, courseRun);
+    const courseRunUrl = `${DiscoveryDataApiService.discoveryBaseUrl}/course_runs/${key}/`;
+    // Remove key param from body for API validation
```
Dillon-Dumesnil: What's the necessity for this?

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -226,10 +228,12 @@ export class BaseEditCourseForm extends React.Component {
     const programOptions = (courseRunOptions &&
       this.parseOptions(courseRunOptions.expected_program_type.choices));
 
+    const disabled = courseInReview || !editable;
+
     let submitState = 'default';
     if (submitting || (courseInfo && courseInfo.isSubmittingEdit)) {
       submitState = 'pending';
-    } else if (pristine) {
+    } else if (pristine || disabled) {
```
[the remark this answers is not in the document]
xnick421: without it, when the legal reviewer makes changes to the form, the disabled course button label changes to 'Save and continue editing'.  we dont want that

src/data/services/DiscoveryDataApiService.js
xnick421: when doing internal update, the api only accepts ofac_resctriction, ofac_comment, and status.  no need to pass key, it wont do anything with it.  will actually throw error for irrelevant params.

PR #237, opened 2019-08-09
Search bar styling and clear status alerts

ciduarte:
Previously, clearing one status alert would clear all
of the status alerts visible. Separated the dismissAlert
logic to allow the alerts to be cleared one at a time.


ciduarte:
FYI, before & after:
Screen Shot 2019-08-09 at 11 37 35 AM
Screen Shot 2019-08-09 at 11 36 41 AM

PR #239, opened 2019-08-12
DISCO-1299 remove date type from date inputs

xnick421:


src/components/DateTimeField/index.jsx
```diff
@@ -74,10 +73,11 @@ class DateTimeField extends React.Component {
               />
             }
             placeholder="mm/dd/yyyy"
+            pattern="(0[1-9]|1[012])[- /.](0[1-9]|[12][0-9]|3[01])[- /.](19|20)\d\d"
```
Dillon-Dumesnil: Has this been tested with international dates where they don't follow the order of MM/DD/YYYY? Like in the past with our international partners, the date field would automatically switch to being in DD/MM/YYYY and they could enter it in that way.

src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -177,6 +179,27 @@ class CollapsibleCourseRun extends React.Component {
     return 'Submit Run for Review';
   }
 
+  formatDate(date) {
+    if (date) {
+      if (date.includes('T') &&
+        date.split('T')[0].match(/(19|20)\d{2}-(0\d{1}|1[0-2])-([0-2]\d{1}|3[0-1])/)) {
```
Dillon-Dumesnil: I think all of these regexes should be written into a constants file with more clear names.

src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -155,7 +154,6 @@ class BaseCreateCourseForm extends React.Component {
             timeLabel={`Start time (${localTimeZone})`}
             helpText={startDateHelp}
```
Dillon-Dumesnil: The help text should probably be updated to specify that the day and month need to be two digits and the year needs to be 4 so users don't run into the situation where "1-1-2020" isn't saveable.

src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -164,7 +162,6 @@ class BaseCreateCourseForm extends React.Component {
             timeLabel={`End time (${localTimeZone})`}
             helpText={endDateHelp}
```
Dillon-Dumesnil: Same here

src/components/DateTimeField/index.jsx
xnick421: @Dillon-Dumesnil I am going to guess this was done by the date-picker component which we were using from paragon.  the only thing we are really relying on here is the place holder to let the user know what the expected input is.  I could add something tho to detect where the user is geographically and then change the logic according to that.  what do you think?

Dillon-Dumesnil:
Have we considered using ```
const isSafari = /constructor/i.test(window.HTMLElement) || (function (p) { return p.toString() === "[object SafariRemoteNotification]"; })(!window['safari'] || (typeof safari !== 'undefined' && safari.pushNotification));
```
as a way of supporting this in Safari without needing to switch the behavior for the rest of our users?

File: src/components/DateTimeField/index.jsx (hunk 1 above)
Comment (Dillon-Dumesnil) [reply src/components/DateTimeField/index.jsx]: Yeah, I believe it was done by default with the date type we were using. I guess I'm hesitant to now have to change the behavior for all of our overseas partners. I think adding in logic about detecting the user geographically would be overkill for this. Just worth noting that we are now hardcoding in the American way of doing dates. Maybe at minimum, add in a comment to revisit this in the future and see if we can go back to using date.

Comment (xnick421) [reply src/components/DateTimeField/index.jsx]: we are switching back to paragon which will support safari, supposedly soon

**Comment by xnick421** [comment]
const isSafari = /constructor/i.test(window.HTMLElement) || (function (p) { return p.toString() === "[object SafariRemoteNotification]"; })(!window['safari'] || (typeof safari !== 'undefined' && safari.pushNotification));     

solutions like this are brittle, in my experience have been known to break with time due to our reliance on variables not declared by us.

**Comment by Dillon-Dumesnil** [comment]
True, but as you said, this is only supposed to be a temporary fix anyway and this way we are able to continue using Paragon for every other browser while Paragon puts in the fix for Safari

**Comment by xnick421** [comment]
Do we know that every version of safari will support this logic?

**Comment by Dillon-Dumesnil** [comment]
According to https://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser, it has been tested in versions `3-10`. I have `12.1.2` running on my machine and it also worked there.

**Comment by ciduarte** [comment]
Another solution might be to check if the date type is even supported, and adjust the date picker to support only typing in the date: https://gomakethings.com/how-to-check-if-a-browser-supports-native-input-date-pickers/

When I quickly looked into this, using a placeholder on the date input field fixed the issue on Safari (https://stackoverflow.com/questions/35682138/html5-date-picker-doesnt-show-on-safari). The only problem was that our date validation automatically fills the field with "Invalid Date" before you can finish typing the date.

Maybe a combination of these two solutions can help us troubleshoot on Safari? ie. Check if date is supported, if it is not supported, replace the date validation with regex?

**Comment by xnick421** [comment]
safari doesn't support date

**Comment by ciduarte** [comment]
Yes, that's the point

File: Makefile [hunk 5]
```diff
@@ -34,4 +34,4 @@ snapshot:
 	docker exec -it edx.pubfe npm run snapshot
 
 lint:
-	docker exec -it edx.pubfe npm run lint && npm run stylelint
+	docker exec -it edx.pubfe npm run lint && docker exec -it edx.pubfe npm run stylelint
```
Dillon-Dumesnil: Nit: I think this is a cleaner way of making this change so we don't have to copy the docker exec -it edx.pubfe
`docker exec -it edx.pubfe bash -c 'npm run lint && npm run stylelint'`

src/utils/index.js
```diff
@@ -4,11 +4,16 @@ import qs from 'query-string';
 
 import history from '../data/history';
 
-const getDateString = date => (date ? moment.utc(date).format('YYYY-MM-DD') : '');
+const dateWithDashes = date => (date ? moment.utc(date).format('YYYY-MM-DD') : '');
+const dateWithSlashes = date => (date ? moment.utc(date).format('YYYY/MM/DD') : '');
```
Dillon-Dumesnil: Is there a reason to have all of this date with slashes if we really just want dashes everywhere? Why not just always use dashes, Safari or otherwise?

src/helpText.jsx
```diff
@@ -1,12 +1,28 @@
...
+import { isSafari } from './utils';
 
 /**
  * This file holds a few jsx constants to share helpText between different components.
  * Please keep it sorted for ease of use.
  */
 
+const publishDateHelp = (
+  <div>
+    {isSafari && <p>Format: Month(2 digits)/Day(2 digits)/Year(4 digits)</p>}
```
Dillon-Dumesnil: I feel like all of these first messages should match across all of the date fields. Also here, you have it written as mm/dd/yyyy, but the placeholder is yyyy/mm/dd. That seems misleading or am I reading this wrong?

xnick421: missed it

src/utils/index.js
xnick421: the date is rendered now on the app with slashes
do you want to change everything to dashes?

src/components/DateTimeField/index.jsx
```diff
@@ -27,15 +27,13 @@ class DateTimeField extends React.Component {
     const {
       time,
     } = this.state;
-    this.setState({ date: value });
```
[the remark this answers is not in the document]
ciduarte: I ran this locally and if I change both the date and time, and hit save, one or the other isn't the value that I changed it to on reload.

src/components/DateTimeField/index.jsx
```diff
@@ -27,15 +27,13 @@ class DateTimeField extends React.Component {
...
     this.concatDateTime(value, time);
   }
 
   updateTime(value) {
     const {
       date,
     } = this.state;
-    this.setState({ time: value });
```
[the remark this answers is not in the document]
xnick421: ah ok now i see where we need this.  it wasnt clear to me why we were setting the state of these.

src/components/DateTimeField/index.jsx
```diff
@@ -114,16 +120,25 @@ DateTimeField.propTypes = {
     value: PropTypes.string,
     name: PropTypes.string,
     onChange: PropTypes.func.isRequired,
+    placeholder: PropTypes.string,
```
Dillon-Dumesnil: This input prop here is coming from the Redux form Field component. From what I can tell, we aren't doing anything with this and you should only need it listed below with the regular props.

src/data/constants/index.js
```diff
@@ -13,6 +13,9 @@ const IN_REVIEW_STATUS = [REVIEW_BY_LEGAL, REVIEW_BY_INTERNAL];
 const POST_REVIEW_STATUSES = [PUBLISHED, REVIEWED];
 
 const DATE_FORMAT = 'YYYY-MM-DD[T]HH:mm:ss[Z]';
+const DATE_INPUT_GUARD = '20[1-9][0-9]/(0[1-9]|1[012])/(0[1-9]|[12][0-9]|3[01])';
```
Dillon-Dumesnil: I think this variable name would make more sense as `DATE_PATTERN` since everywhere it is used, you are passing it in as the pattern prop.

src/helpText.jsx
```diff
@@ -1,12 +1,28 @@
...
+import { isSafari } from './utils';
 
 /**
  * This file holds a few jsx constants to share helpText between different components.
  * Please keep it sorted for ease of use.
  */
 
+const publishDateHelp = (
+  <div>
+    { isSafari && <p>Required Format: YYYY/MM/DD</p> }
```
Dillon-Dumesnil: Nit: I would also have this match the case that we show in the placeholder. So either uppercase in both or lowercase in both.

PR #240, opened 2019-08-13
Show PCs on course edit page

mikix:
https://openedx.atlassian.net/browse/DISCO-1269


mikix:
@ciduarte I don't think it's a real use case for edx.org - maybe for open edx users though. But I felt like not showing anything at all was a nicer UX, on the assumption that it was a fine state, not something we should try to get the user to fix. But maybe that's wrong? I dunno. I don't think edx.org will hit it anyway.

PR #243, opened 2019-08-19
DISCO-1307 status change exempt

xnick421:
change notification message when a reviewed run stays in review after a change.  Change button to disabled when no changes are made, and change label according to if exempt or non-exempt fields are changed

### DIFF
--- a/src/components/EditCoursePage/CollapsibleCourseRun.jsx
+++ b/src/components/EditCoursePage/CollapsibleCourseRun.jsx
@@ -8,7 +8,8 @@ import ActionButton from '../ActionButton';
 import ButtonToolbar from '../ButtonToolbar';
 import { courseSubmittingInfo } from '../../data/actions/courseSubmitInfo';
 import FieldLabel from '../FieldLabel';
-import { localTimeZone, formatDate, isSafari, getDateWithDashes, getDateWithSlashes } from '../../utils/index';
+import { localTimeZone, formatDate, isSafari, getDateWithDashes, getDateWithSlashes,
+  isNonExemptChanged, isPristine } from '../../utils/index';
 import Pill from '../Pill';
 import RenderInputTextField from '../RenderInputTextField';
 import RenderSelectField from '../RenderSelectField';
@@ -19,7 +20,7 @@ import TranscriptLanguage from './TranscriptLanguage';
 
 import {
   DATE_FORMAT, IN_REVIEW_STATUS, REVIEW_BY_INTERNAL, REVIEW_BY_LEGAL,
-  PUBLISHED, DATE_INPUT_PATTERN, FORMAT_DATE_MATCHER, NORMALIZE_DATE_MATCHER,
+  PUBLISHED, DATE_INPUT_PATTERN, FORMAT_DATE_MATCHER, NORMALIZE_DATE_MATCHER, REVIEWED,
 } from '../../data/constants';
 import { endDateHelp, startDateHelp, pacingHelp, publishDateHelp } from '../../helpText';
 import RichEditor from '../RichEditor';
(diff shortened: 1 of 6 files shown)


src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -175,8 +191,12 @@ class CollapsibleCourseRun extends React.Component {
           break;
       }
     }
-    if (status === PUBLISHED) {
-      return 'Publish Run';
+    if (status === PUBLISHED) return 'Publish Run';
```
jlajoie: nit: get outta here with these 1 liners without brackets, I'll swap the linter if not 😁

src/utils/index.js
```diff
@@ -101,6 +102,13 @@ const getErrorMessages = (error) => {
   return ['Unknown error.'];
 };
 
+const isNonExemptChanged = (initialRuns, currentRuns, runKey) => {
```
jlajoie: I always feel weird about method names with negations, but this is fine, just is a bit hard to read when you go through something like ```if not isNonExemptChanged```

src/components/EditCoursePage/index.jsx
```diff
@@ -256,12 +288,20 @@ class EditCoursePage extends React.Component {
...
+    const { targetRun: { key } } = this.state;
+    const runFromAPI = course_runs ? course_runs.find(run => run.key === key) : {};
     switch (status) {
       case REVIEW_BY_LEGAL:
         return 'Legal Review Complete. Course Run is now awaiting PC Review.';
       case REVIEW_BY_INTERNAL:
         return 'PC Review Complete.';
       default:
+        if (status === REVIEWED && runFromAPI.status === REVIEWED) {
+          // if run was in reviewed state when changed and stays in review state, show this message
```
Dillon-Dumesnil: This comment doesn't add value

src/components/EditCoursePage/index.jsx
```diff
@@ -256,12 +288,20 @@ class EditCoursePage extends React.Component {
...
+          return 'Course Run Updated.';
+        }
+        // if run was in reviewed state when changed and now needs legal review, show this message
```
Dillon-Dumesnil: Same here.

src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -663,6 +696,15 @@ CollapsibleCourseRun.defaultProps = {
   targetRun: null,
   isOpen: false,
   authentication: {},
+  currentFormValues: {
```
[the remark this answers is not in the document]
xnick421: this is because the course_runs property isn't always present, so we give it a default value of [], instead of checking if its present in the code

src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -178,6 +196,13 @@ class CollapsibleCourseRun extends React.Component {
     if (status === PUBLISHED) {
       return 'Publish Run';
     }
+    if (status === REVIEWED) {
+      if (initialRuns && currentRuns) {
+        return isNonExemptChanged(initialRuns, currentRuns, key) ||
```
Dillon-Dumesnil: Nit: Change `initialRuns` -> `initialValues.course_runs` and `currentRuns` -> `currentFormValues.course_runs`

It would at least let us lose those six lines above from deconstructing now that we already have to parse out currentFormValues and initialValues

src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -164,8 +165,25 @@ class CollapsibleCourseRun extends React.Component {
     }
   }
 
-  displayDefaultButtonLabel(status, isAdmin) {
```
Dillon-Dumesnil: nit: Leave passing in isAdmin since we already pull out administrator inside of render and it can avoid the three lines here.

PR #248, opened 2019-08-28
Fix staff submission error clearing

mikix:
If you created a new staff member after already having a staff submission error, sometimes that error would hang around. Now it is cleared correctly.

https://openedx.atlassian.net/browse/DISCO-1312


src/utils/validation.js
```diff
@@ -109,6 +111,9 @@ const editCourseValidate = (values, props) => {
           }
         }
       });
+      if (Object.keys(runErrors).length > 0) {
```
xnick421: `Object.keys(runErrors).length` would be fine.  if length is 0 it evaluates to false anyway

PR #253, opened 2019-08-30
Give more help on dates

mikix:
Per feedback from Carol


src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -221,6 +221,23 @@ class CollapsibleCourseRun extends React.Component {
     return '';
   }
 
+  dateEditHelp(courseRun) {
```
jlajoie: Might be worth moving to `src/helpText.jsx` to put it with the other date helpers?

mikix: Yeah, you're right. I kept it out because it relied on courseRun argument, unlike the others there. But that's no real reason - better to consolidate.

mikix: Done.

PR #254, opened 2019-09-05
Avoid submission nested error loops

mikix:
Stop passing down a validate method for the whole form down to the
field array of runs. This method was causing a false positive on
errors being present.

### DIFF
--- a/src/components/EditCoursePage/EditCourseForm.jsx
+++ b/src/components/EditCoursePage/EditCourseForm.jsx
@@ -69,7 +69,7 @@ export class BaseEditCourseForm extends React.Component {
     const stoppingRunReview = prevProps.courseSubmitInfo.isSubmittingRunReview &&
                               !courseSubmitInfo.isSubmittingRunReview;
     const hasCourseErrors = courseSubmitInfo.errors &&
-                            courseSubmitInfo.errors !== {} &&
+                            Object.keys(courseSubmitInfo.errors).length &&
                             Object.keys(courseSubmitInfo.errors) !== ['course_runs'];
     if (stoppingRunReview && hasCourseErrors) {
       this.openCollapsible();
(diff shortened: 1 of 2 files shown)


src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -69,7 +69,7 @@ export class BaseEditCourseForm extends React.Component {
     const stoppingRunReview = prevProps.courseSubmitInfo.isSubmittingRunReview &&
                               !courseSubmitInfo.isSubmittingRunReview;
     const hasCourseErrors = courseSubmitInfo.errors &&
-                            courseSubmitInfo.errors !== {} &&
+                            Object.keys(courseSubmitInfo.errors).length &&
```
mikix: Unrelated change - you can't compare objects like they are dicts, because dicts aren't a real thing in JS.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -845,6 +845,7 @@ export class BaseEditCourseForm extends React.Component {
             collapsiblesOpen={this.state.collapsiblesOpen}
             onToggle={(index, value) => this.toggleCourseRun(index, value)}
             {...this.props}
+            validate={() => {}} // override method from our props, we don't want to pass it down
```
mikix: This does not prevent validation from occuring inside runs. I tested both course fields and run fields (manually checked and html5-validated for each).

mikix: To fully explain:

So redux was coming up with errors from the form, but they weren't real, it was like an errors object like { course_runs: { _error: {} } } or something
But that tricks redux, despite not being a real error
It expects just {} on all-good
And it seemed to be because we were passing down the top level validate prop into the FieldArray for the runs
So I stopped passing it down by overriding with a dummy validate
And yet, somehow the fields in the FieldArray are still validated...
I don't fully understand it, tbh

PR #255, opened 2019-09-06
Rename publisher-frontend to frontend-app-publisher

nasthagiri:
Related PRs:

- https://github.com/edx/edx-gomatic/pull/576
- https://github.com/edx/course-discovery/pull/2142
- https://github.com/edx/terraform/pull/1852

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# publisher-frontend
+# frontend-app-publisher
 
-[![Greenkeeper badge](https://badges.greenkeeper.io/edx/publisher-frontend.svg)](https://greenkeeper.io/)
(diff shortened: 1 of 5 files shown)


nasthagiri:
Question to reviewers: Should the image name in the [docker-compose file](https://github.com/edx/frontend-app-publisher/blob/3976706f7f71685e4aabe118ec3278836159797e/docker-compose.yml#L10) also be updated?  That may interrupt current development.

nasthagiri:
@mikix Thanks for the tip about DockerHub alias.  I've created an Ops ticket to do that work since I don't have access there.

PR #258, opened 2019-09-12
Have pacing type be read-only on edits

mikix:
We're moving that source-of-edit-truth to Studio. You still set the pacing type on run creation in Publisher. But edit it only in Studio.

https://openedx.atlassian.net/browse/DISCO-1341

This is related to https://github.com/edx/course-discovery/pull/2151


mikix:
Yeah, I forgot to remove them in a previous PR...

PR #287, opened 2019-09-23
DISCO-1371: Only save course runs that changed

Dillon-Dumesnil:
To prevent load on discovery, help with requests/min, and help avoid
deadlock. This will look for course runs that were changed
as well as saving non-archived course runs if the course mode or price
changed.

### DIFF
--- a/src/components/EditCoursePage/index.jsx
+++ b/src/components/EditCoursePage/index.jsx
@@ -7,7 +7,7 @@ import EditCourseForm from './EditCourseForm';
 import PageContainer from '../PageContainer';
 import StatusAlert from '../StatusAlert';
 import LoadingSpinner from '../LoadingSpinner';
-import { getCourseNumber, isValidDate, isNonExemptChanged } from '../../utils';
+import { courseRunIsArchived, getCourseNumber, isValidDate, isNonExemptChanged } from '../../utils';
 import { IN_REVIEW_STATUS, PUBLISHED, REVIEW_BY_INTERNAL, REVIEW_BY_LEGAL, REVIEWED,
   UNPUBLISHED } from '../../data/constants';
 import ConfirmationModal from '../ConfirmationModal';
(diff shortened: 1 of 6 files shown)


src/components/EditCoursePage/index.jsx
```diff
@@ -55,11 +56,40 @@ class EditCoursePage extends React.Component {
...
+        (
+          // Only send runs if they have changed OR the course mode or price has
+          // changed and the run is "archived"
+          JSON.stringify(initialCourseRunValues[i]) !== JSON.stringify(run) ||
+          (
+            (initialMode !== courseData.mode || initialPrice !== courseData.price) &&
+            !(run.status === 'unpublished' && moment().isAfter(run.end))
+          )
+        )
+      )
```
mikix: As we talked about in person, this is getting a little hard to follow.

Maybe use an early exit strategy?
```
// explanation...
if (targetRun && (run.key === targetRun.key)) {
  return true;
}

// explanation 2...
if ()...
```

Or create a bunch of intermediate variables:
```
const targetted = targetRun && (run.key === targetRun.key);
const reviewing = IN_REVIEW_STATUS.includes(run.status);
return targetted || (!reviewing...);
```

src/components/EditCoursePage/index.jsx
```diff
@@ -55,11 +56,40 @@ class EditCoursePage extends React.Component {
...
+    const modifiedCourseRuns = courseData.course_runs.filter((run, i) => (
+      // If we are submitting a run for review or re-publishing a run, it should
+      // always get through to the backend
+      (targetRun && (run.key === targetRun.key)) ||
+      // Don't send any courses in review - backend will reject them
+      (
+        !IN_REVIEW_STATUS.includes(run.status) &&
+        (
+          // Only send runs if they have changed OR the course mode or price has
+          // changed and the run is "archived"
```
mikix: Maybe just for the dummies, say *why* we care about mode and price ("they are used on the backend to determine course run level info" or something).

src/components/EditCoursePage/index.jsx
```diff
@@ -55,11 +56,40 @@ class EditCoursePage extends React.Component {
...
+          JSON.stringify(initialCourseRunValues[i]) !== JSON.stringify(run) ||
+          (
+            (initialMode !== courseData.mode || initialPrice !== courseData.price) &&
+            !(run.status === 'unpublished' && moment().isAfter(run.end))
```
mikix: Do we not have a utliity method or something for that?

PR #288, opened 2019-09-24
Prefer key_for_reruns as course number

mikix:
Whenever we show the course number, prefer key_for_reruns, as that is what the course team member is going to be thinking of as their key.

This doesn't change how we use course keys, just how we display them.

https://openedx.atlassian.net/browse/DISCO-1358


mikix:
I didn't change any field labels? I only added the `number` one because now we have both `key` and `key_for_reruns`

jlajoie:
> I didn't change any field labels? I only added the `number` one because now we have both `key` and `key_for_reruns`

From "key" to "number". I guess "number" _does_ make a bit more sense since it's not going to be just the de-facto key anymore.

PR #289, opened 2019-09-26
Compare objects with stable stringify

mikix:
Add dependency on fast-json-stable-stringify (it was already in our dependency tree) to deterministically compare objects.

Hopefully will fix us accidentally sending too many course runs when saving and hitting deadlocks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "connected-react-router": "5.0.1",
     "email-prop-type": "3.0.0",
     "eslint": "5.16.0",
+    "fast-json-stable-stringify": "2.0.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -97,7 +98,7 @@
     "style-loader": "0.23.1",
     "webpack": "4.41.0",
     "webpack-cli": "3.3.9",
-    "webpack-dev-server": "3.2.1",
+    "webpack-dev-server": "3.8.1",
```
mikix: This line was simply to avoid a local devstack problem with newer webpack-cli not shipping a binary dev-server needed.

PR #290, opened 2019-09-27
Avoid calling forEach on an object

mikix:
This will avoid an unnecessary error when PCs and Legal submit internal reviews.

### DIFF
--- a/src/components/EditCoursePage/index.jsx
+++ b/src/components/EditCoursePage/index.jsx
@@ -260,7 +260,7 @@ class EditCoursePage extends React.Component {
       this.prepareSendCourseRunData(courseData);
     // Process courseData to reduced data set
     const courseEditData = this.prepareSendCourseData(courseData, modifiedCourseRuns);
-    return editCourse(courseEditData, modifiedCourseRuns, !!targetRun);
+    return editCourse(courseEditData, modifiedCourseRuns, !!targetRun, !!isInternalReview);
   }
 
   displayReviewStatusAlert(status) {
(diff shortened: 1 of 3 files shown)


src/data/actions/courseInfo.js
```diff
@@ -184,16 +178,14 @@ function createCourse(courseData) {
...
-    ApiFunctionEnum.EDIT_RUNS_FUNCTION;
+function handleCourseRuns(dispatch, courseRunData, course, submitReview, internalReview) {
   // make course copy so we are not re-assigning properties of this functions original params
   const newCourse = Object.assign({}, course);
-  DiscoveryDataApiService[functionCall](courseRunData).then((runResponse) => {
+  let sendData = DiscoveryDataApiService.editCourseRuns;
+  if (internalReview) {
+    sendData = data => Promise.all([DiscoveryDataApiService.internalReviewEdit(data)]);
+  }
+  sendData(courseRunData).then((runResponse) => {
```
mikix: I think this whole arrangement is _cleaner_, but not ideal. This layer shouldn't know about details like this. Ideally `handleCourseSubmit` would either pass down this `sendData` callback itself or really, start a different thread than `editCourse` which mostly assumes its dealing with lists.

But... this fix is slightly better and fairly easy. So I did this.

src/components/EditCoursePage/index.jsx
```diff
@@ -260,7 +260,7 @@ class EditCoursePage extends React.Component {
       this.prepareSendCourseRunData(courseData);
     // Process courseData to reduced data set
     const courseEditData = this.prepareSendCourseData(courseData, modifiedCourseRuns);
-    return editCourse(courseEditData, modifiedCourseRuns, !!targetRun);
+    return editCourse(courseEditData, modifiedCourseRuns, !!targetRun, !!isInternalReview);
```
jlajoie: You shouldn't need the `!!` here since it's already a boolean

PR #297, opened 2019-09-30
DISCO-1372: Set title on loading dashboard

Dillon-Dumesnil:


src/components/CourseTable/index.jsx
```diff
@@ -86,6 +87,9 @@ class CourseTable extends React.Component {
         {showDashboard &&
         (
           <React.Fragment>
+            <Helmet>
+              <title>Publisher Beta | edX</title>
```
mikix: `edX`?!

PR #304, opened 2019-10-10
DISCO-1388 course run redirect component

xnick421:


src/data/services/DiscoveryDataApiService.js
```diff
@@ -17,6 +17,19 @@ class DiscoveryDataApiService {
     });
   }
 
+  static fetchCourseRuns(params) {
```
Dillon-Dumesnil: This feels like a weird function name when it's returning a single course_uuid. You aren't fetching any course runs. Maybe `fetchCourseThroughCourseRun` or something like that, but `fetchCourseRuns` feels weird.

Dillon-Dumesnil: Also shouldn't this more closely follow the above method? We are only sending in a single course run, right? Why not make the path `${DiscoveryDataApiService.discoveryBaseUrl}/course_runs/${key}`?

src/components/CourseRunRedirect/index.jsx
```diff
@@ -0,0 +1,33 @@
...
+        .fetchCourseRuns({ keys: this.props.courseRunKey });
+      const uuid = response.data.results[0].course_uuid;
+      this.setState({ path: `/courses/${uuid}` }); // eslint-disable-line react/no-did-mount-set-state
+    } catch (e) {
+      this.setState({ path: '/' }); // eslint-disable-line react/no-did-mount-set-state
+    }
+  }
+
+  render() {
+    return this.state.path ? <Redirect to={this.state.path} /> : null;
```
Dillon-Dumesnil: This probably shouldn't be rendering null and instead render the `LoadingSpinner` component so it is consistent with actually loading a course.

src/data/services/DiscoveryDataApiService.js
xnick421: Your totally right, didn't realize the course run detail accepted a key

PR #311, opened 2019-10-16
DISCO-1397 handle 404 from org users endpoint

xnick421:
Allow 404 to bypass Promise.all catch so that
we gracefully display available users


src/components/PageContainer/PageContainer.scss
```diff
@@ -1,3 +1,3 @@
 .content-container {
   min-height: calc(100vh - 300px);
-}
```
jlajoie: nit: don't need this change here

src/components/SidePanes/UsersPane.jsx
```diff
@@ -119,6 +121,7 @@ class UsersPane extends React.Component {
             ))}
           </div>
         }
+        {!showEditors && organizationUsers.error && <div className="error-msg">{organizationUsers.error[0]}</div>}
```
jlajoie: We have a ```StatusAlert``` component I think we've used elsewhere for showing errors, might be worth just using that guy instead.

src/components/SidePanes/UsersPane.scss
```diff
@@ -0,0 +1,3 @@
+.error-msg {
+  color: #FF0000;
```
jlajoie: nit: I think typically we've just used bootstrap's color class names, but this whole file might go away if you swap to StatusAlert

src/components/SidePanes/UsersPane.jsx
jlajoie: Also might be worth checking if the len of organizationUsers.error > 0 as well just for safety

src/components/SidePanes/UsersPane.scss
xnick421: ty on headsup with status alert component

PR #313, opened 2019-10-18
feat: upgrade paragon

abutterworth:
Upgrade Paragon, remove edx-bootstrap, and update the prop usage of some components.

Screen Shot 2019-10-18 at 2 53 52 PM
Screen Shot 2019-10-18 at 2 53 57 PM
Screen Shot 2019-10-18 at 2 53 48 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,11 +18,10 @@
   },
   "license": "AGPL-3.0",
   "dependencies": {
-    "@edx/edx-bootstrap": "1.0.4",
     "@edx/frontend-auth": "5.3.6",
     "@edx/frontend-component-footer": "4.1.5",
     "@edx/frontend-logging": "2.1.0",
-    "@edx/paragon": "^7.1.5",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -22,7 +22,7 @@
     "@edx/frontend-auth": "5.3.6",
     "@edx/frontend-component-footer": "4.1.5",
     "@edx/frontend-logging": "2.1.0",
-    "@edx/paragon": "4.2.4",
+    "@edx/paragon": "^7.1.5",
```
mikix: Should we still be specifying edx-bootstrap manually like we do?

mikix: (Also, the lack of `^` carots is because we use renovate on the repo, and it hardcodes all the versions. So might want to keep that going.)

abutterworth: No we should remove, I'll update and pin the paragon version

PR #315, opened 2019-10-21
Adding support for Type in Publisher

Dillon-Dumesnil:
DISCO-1386


src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -121,32 +134,64 @@ class BaseCreateCourseForm extends React.Component {
...
+              <Field
+                name="type"
+                component={RenderSelectField}
+                options={courseTypeOptions}
+                label={
+                  <FieldLabel
+                    id="course-type-label"
+                    text="Course Type"
+                    required
+                    helpText={(<div><p>TODO: Come up with type helpText</p></div>)}
```
mikix: Ahem. (Same TODO on runs too)

src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -121,32 +134,64 @@ class BaseCreateCourseForm extends React.Component {
...
+                  min: 1.00,
+                  step: 0.01,
+                  max: 10000.00,
+                }}
+                type="number"
+                label={<FieldLabel text="Price (USD)" required />}
+                required
+              />}
+            </React.Fragment>) : (
+              <React.Fragment>
```
mikix: Super-nit: different indentation levels for these two is odd to me.

src/components/CreateCoursePage/CreateCourseForm.jsx
```diff
@@ -121,32 +134,64 @@ class BaseCreateCourseForm extends React.Component {
...
+          {usingCourseType ? (
+            <React.Fragment>
+              <Field
+                name="type"
+                component={RenderSelectField}
+                options={courseTypeOptions}
+                label={
+                  <FieldLabel
+                    id="course-type-label"
+                    text="Course Type"
```
mikix: I know to us engineers, this is now a "Course Type" but to the user, nothing has changed, right? Might want to keep labelling it as "Enrollment track"?  (also those two strings have different capitalization schemes)

I guess you have to differentiate the course from the run. But I don't know what language users would be comfortable here. "Course type" is not a thing I think they've seen before. I guess I'm saying this might be a good question for Product.

src/components/CreateCoursePage/CreateCourseForm.test.jsx
```diff
@@ -53,8 +37,21 @@ describe('CreateCourseForm', () => {
...
       organizations={organizations}
-      getCourseRunOptions={() => courseRunOptionsReturn}
-      parseOptions={() => parseOptionsReturn}
+      courseOptions={courseOptions}
+      courseRunOptions={courseRunOptions}
+    />);
+    expect(component).toMatchSnapshot();
+  });
+
+  it('renders html correctly with data and using Course Type', () => {
```
mikix: Booo snapshot tests. But fine.

src/utils/index.js
```diff
@@ -155,6 +155,48 @@ const isPristine = (initialValues, currentFormValues, runKey) => {
...
+      const courseRunType = type.course_run_types[j];
+      courseRunTypeArray.push({
+        label: courseRunType.name, value: courseRunType.uuid,
+      });
+    }
+    courseRunTypeOptions[type.uuid] = courseRunTypeArray;
+    // This should be changed to remove the part about AUDIT_TRACK once we stop
+    // creating draft Audit entitlements. Right now, this is used to decide if we
+    // show the price on the course create/edit form.
+    if (type.entitlement_types.length && type.entitlement_types[0] !== AUDIT_TRACK.key) {
```
mikix: `[0]` seems brittle. Maybe also add `length == 1` instead of just `>0`? But I get that you are hoping for a short lifetime for this code.

src/data/constants/testData.js
```diff
@@ -0,0 +1,164 @@
...
+        type: {
+          type: 'field',
+          required: false,
+          read_only: false,
+          label: 'Type',
+          type_options: [
+            {
+              course_run_types: [
+                {
+                  uuid: 'f17e29d6-4648-4bb5-a199-97dc40f904aa',
```
mikix: All these real UUIDs -- did you grab devstack output?

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -203,30 +172,41 @@ export class BaseEditCourseForm extends React.Component {
...
+      this.parseOptions(courseRunOptionsData.pacing_type.choices));
+    const languageOptions = (courseRunOptionsData &&
+      this.parseOptions(courseRunOptionsData.content_language.choices));
+    const programOptions = (courseRunOptionsData &&
+      this.parseOptions(courseRunOptionsData.expected_program_type.choices));
+    const parsedTypeOptions = courseOptionsData &&
+      parseCourseTypeOptions(courseOptionsData.type.type_options);
+    const { courseTypeOptions } = parsedTypeOptions;
+    const { courseRunTypeOptions } = parsedTypeOptions;
+    const { entitlementUUIDS } = parsedTypeOptions;
```
mikix: The linter allowed this? I thought our style was
```
    const {
        courseTypeOptions,
        courseRunTypeOptions,
        entitlementUUIDS,
    } = parsedTypeOptions;
```

That said, your style is less lines, but more characters.... hmmm

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -302,35 +282,70 @@ export class BaseEditCourseForm extends React.Component {
...
-              name="mode"
-              component={RenderSelectField}
-              label={
-                <FieldLabel
-                  id="mode.label"
-                  text="Enrollment track"
-                  helpText={enrollmentHelp}
-                  extraText="Cannot edit after submission"
+            {type ? (
+              <React.Fragment>
```
mikix: Should we be sharing more code between the edit form and the create form? Seems like we have a lot of dup'd lines.

src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -355,6 +351,24 @@ class CollapsibleCourseRun extends React.Component {
...
+            options={currentFormValues.type ? courseRunTypeOptions[currentFormValues.type] : [{ label: 'Select Course Type first', value: '' }]}
+            extraInput={{ onInvalid: this.openCollapsible }}
+            label={
+              <FieldLabel
+                id="run_type.label"
+                text="Course Run Type"
+                helpText={(<div><p>TODO: Come up with run type helpText</p></div>)}
+              />
+            }
+            // Should this be disabled in other scenarios?
```
mikix: Is this a comment we should address now? You can't edit after submission, right? Do we need that extraHelp here with a disable in that case?

src/components/CreateCourseRunPage/index.jsx
```diff
@@ -29,37 +27,32 @@ class CreateCourseRunPage extends React.Component {
...
       courseInfo,
+      courseOptions,
+      courseRunOptions,
+      fetchCourseInfo,
+      fetchCourseOptions,
+      fetchCourseRunOptions,
       id,
     } = this.props;
-    if (!courseInfo ||
+    if (!!courseInfo ||
```
mikix: You're inverting this check? I dunno if that makes sense, since `{}` is always truthy, so this bit isn't doing anything useful anymore. You can probably just drop it altogether and just have the `entries.length` check. Same below.

src/components/CreateCoursePage/index.jsx
```diff
@@ -183,9 +166,11 @@ class CreateCoursePage extends React.Component {
                 initialValues={initialValues}
                 currentFormValues={formValues}
                 organizations={organizations}
+                // TODO: Add in logic here to decide when course type is being used
```
mikix: Nice. I 100% agree with your delaying this decision. But I do like the idea of like "eh... we should like, add logic maybe?"

src/components/CreateCourseRunPage/CreateCourseRunForm.jsx
```diff
@@ -11,146 +11,179 @@ import ActionButton from '../ActionButton';
...
-                  </p>
-                </div>
-              }
-            />
+
+class BaseCreateCourseRunForm extends React.Component {
+  parseOptions(inChoices) {
+    return inChoices.map(choice =>
+      ({ label: choice.display_name, value: choice.value }));
+  }
```
mikix: Is this worth moving into the `utils.js` file?

src/components/CreateCoursePage/CreateCourseForm.jsx
Dillon-Dumesnil: eslint :(

src/components/CreateCoursePage/CreateCourseForm.jsx
Dillon-Dumesnil: Leaving these as TODO's until we can demo/show this to Product and come up with useful help text together.

src/components/CreateCoursePage/CreateCourseForm.jsx
Dillon-Dumesnil: I can add in a TODO here as well to mark it as a standing product question. Similar to help text, I'm not sure what the best language is yet.

src/data/constants/testData.js
Dillon-Dumesnil: Yeah, they are from my local options response. Is that bad? Should I dummy them out to something else?

src/components/EditCoursePage/EditCourseForm.jsx
Dillon-Dumesnil: It did allow it. Maybe it is better to deconstruct again though for consistency. I don't really have a preference. I'll see if one speaks to me

src/components/EditCoursePage/EditCourseForm.jsx
Dillon-Dumesnil: I'd have to look back into it, but if I remember correctly, there are some small differences between the two forms that make it hard to reuse code.

src/data/constants/testData.js
mikix: Naw, I just thought it was funny.

src/components/CreateCourseRunPage/index.jsx
Dillon-Dumesnil: Right. It used to default to `null` so I think the `!courseInfo` handled it. I've updated the logic to check for `courseInfo.data` since leaving just the `entries.length` check fails if the object does not have `courseInfo.data` defined (and our default value is `{}`). Alternatively, might be able to default to `courseInfo = { data: {} }`, but I'm not sure what is better

src/components/EditCoursePage/CollapsibleCourseRun.jsx
Dillon-Dumesnil: Talked about in person. We believe it should be be disabled as soon as any of the seats have a sku. Just need to write the filter function for that.

src/utils/index.js
```diff
@@ -155,6 +155,52 @@ const isPristine = (initialValues, currentFormValues, runKey) => {
...
+    return [];
+  }
+  const { data } = options;
+  if (!data || !data.actions) {
+    return [];
+  }
+  return data.actions.POST;
+};
+
+const parseCourseTypeOptions = (typeOptions) => {
```
xnick421: const parseCourseTypeOptions = (typeOptions) => {
  const courseRunTypeOptions = {};
  const entitlementUUIDS = [];
  const courseTypeOptions = [{ label: 'Select enrollment track', value: '' }].concat(
  	typeOptions.map((type) => {
  	  courseRunTypeOptions[type.uuid] = [{ label: 'Select enrollment track', value: '' }].concat(
	    type.course_run_types.map(courseRunType => ({ abel: courseRunType.name, value: courseRunType.uuid }))
      );
      // This should be changed to remove the part about AUDIT_TRACK once we stop
	  // creating draft Audit entitlements. Right now, this is used to decide if we
	  // show the price on the course create/edit form.
	  if (type.entitlement_types.length && !(type.entitlement_types.length === 1 && type.entitlement_types[0] === AUDIT_TRACK.key)) {
	    entitlementUUIDS.push(type.uuid);
	  }
  	  return { label: type.name, value: type.uuid };
  	})
  );
  return {
    entitlementUUIDS,
    courseRunTypeOptions,
    courseTypeOptions,
  };
};

xnick421: something like this may be simpler with map and concat

Dillon-Dumesnil: Implemented using the above logic. Thanks Nick!

Dillon-Dumesnil: Implemented using the above logic. Thank Nick!

PR #321, opened 2019-10-29
Adding external key into Publisher UI

Dillon-Dumesnil:
This field will only be displayed if the selected Course Run Type
has a masters mode slug associated with it.

DISCO-1092

### DIFF
--- a/src/components/CreateCourseRunPage/index.jsx
+++ b/src/components/CreateCourseRunPage/index.jsx
@@ -72,6 +72,7 @@ class CreateCourseRunPage extends React.Component {
       end: options.end,
       pacing_type: options.pacing_type,
       rerun: options.rerun,
+      run_type: options.run_type,
     };
     return createCourseRun(uuid, courseRunData);
   }
--- a/src/components/EditCoursePage/CollapsibleCourseRun.jsx
+++ b/src/components/EditCoursePage/CollapsibleCourseRun.jsx
@@ -9,7 +9,7 @@ import ButtonToolbar from '../ButtonToolbar';
 import { courseSubmittingInfo } from '../../data/actions/courseSubmitInfo';
 import FieldLabel from '../FieldLabel';
 import { courseRunIsArchived, localTimeZone, formatDate, isSafari, getDateWithDashes,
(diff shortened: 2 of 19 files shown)


src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -133,6 +138,18 @@ class CollapsibleCourseRun extends React.Component {
       const element = document.getElementById(`${courseId}.staff.label`);
       element.scrollIntoView({ behavior: 'smooth', block: 'start', inline: 'nearest' });
     }
+
+    // We initialize hasExternalKey to null and then the function will reevaluate it into
+    // a boolean which will be our exit state
+    if (prevState.hasExternalKey === null) {
+      const hasExternalKey = currentFormValues.course_runs &&
```
mikix: Shouldn't the `currentFormValues.course_runs` check be part of the `if`?

src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -554,6 +574,30 @@ class CollapsibleCourseRun extends React.Component {
...
+            type="text"
+            label={
+              <FieldLabel
+                id={`${courseId}.external_key.label`}
+                text="Institution Course ID"
+                helpText={
+                  <div>
+                    <p>
+                      Only fill in this field if you are planning to integrate with your
+                      Institution&apos;s systems through edX registrar service.
```
mikix: `&apos` is the straight apostrophe - just use the `’` character (curly apostrophe) directly maybe?

src/components/CreateCourseRunPage/index.jsx
```diff
@@ -72,6 +72,7 @@ class CreateCourseRunPage extends React.Component {
       end: options.end,
       pacing_type: options.pacing_type,
       rerun: options.rerun,
+      run_type: options.run_type,
```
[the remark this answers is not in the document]
Dillon-Dumesnil: I don't think this is that scenario. `options` here is the resulting data from clicking create on the create course run page.

PR #322, opened 2019-10-30
Switch org whitelist to a blacklist

jmyatt:


src/components/CourseTable/index.jsx
```diff
@@ -108,12 +108,12 @@ class CourseTable extends React.Component {
 
   isOrgWhitelisted() {
     const userOrgs = this.props.publisherUserInfo.organizations;
-    if (!orgWhitelist || (orgWhitelist && orgWhitelist.length === 0)) {
-      // No Whitelist specified allow all orgs
+    if (!orgBlacklist || (orgBlacklist && orgBlacklist.length === 0)) {
+      // No Blacklist specified allow all orgs
       return true;
     }
     for (let i = 0; i < userOrgs.length; i += 1) {
```
xnick421: I think the following would be better:

`return userOrgs.any(org => !orgBlacklist.includes(org.key))`

something like this

jmyatt:
@Dillon-Dumesnil about coordinating this stuff, the plan is:
- deploy config that adds the new front end and back end config variables (leaving the existing ones alone)
- deploy related discovery PR to rely on the old publisher whitelist
- deploy this PR to rely on new publisher blacklist

PR #335, opened 2019-11-07
DISCO-1447 refresh data after review

xnick421:


src/data/actions/courseInfo.js
```diff
@@ -192,22 +199,40 @@ function handleCourseRuns(dispatch, courseRunData, course, submitReview, interna
...
   });
 }
 
-function editCourse(courseData, courseRunData, submittingRunForReview, isInternalReview) {
+function editCourse(
+  courseData,
+  courseRunData,
+  submittingRunForReview,
+  isInternalReview,
+  refreshData,
```
Dillon-Dumesnil: Do all of these functions have to be so many lines like this? Is there a way to do it in only 2 or 3 lines?

src/components/EditCoursePage/EditCoursePage.test.jsx
```diff
@@ -576,6 +576,7 @@ describe('EditCoursePage', () => {
       component.setState({
         submitConfirmVisible: true,
       });
+      component.instance().getData = jest.fn();
```
Dillon-Dumesnil: Is there a way to just mock this out once at the top and it holds true for all future calls? I thought we might do something similar for mocking out calls to `now()` or something like that.

PR #336, opened 2019-11-07
Default run type value based on course type selection

Dillon-Dumesnil:
DISCO-1444


src/utils/index.js
```diff
@@ -179,10 +179,10 @@ const parseCourseTypeOptions = (typeOptions) => {
   const runTypeModes = {};
   const entitlementUUIDS = [];
   const courseTypeOptions = [{ label: 'Select enrollment track', value: '' }].concat(typeOptions.map((type) => {
-    courseRunTypeOptions[type.uuid] = [{ label: 'Select enrollment track', value: '' }].concat(type.course_run_types.map((courseRunType) => {
+    courseRunTypeOptions[type.uuid] = type.course_run_types.map((courseRunType) => {
```
mikix: Does this mean that when you switch course type, you now get the default run type instead of having to select one? Is that good or bad?

src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -138,11 +143,14 @@ class CollapsibleCourseRun extends React.Component {
       element.scrollIntoView({ behavior: 'smooth', block: 'start', inline: 'nearest' });
     }
 
+    if (currentType !== prevProps.type) {
+      change(`${courseId}.run_type`, courseRunTypeOptions[currentType][0].value);
+    }
```
mikix: I'm not familiar with what this will do in redux land.

src/utils/index.js
Dillon-Dumesnil: It does. I figured if we want default value, there isn't really a point in forcing them back to a value that they have to change. Unless we only wanted to support default value on create pages and the edit still goes back to having to select one?

src/components/EditCoursePage/CollapsibleCourseRun.jsx
Dillon-Dumesnil: This is just going to dispatch the `CHANGE` action to update the current form value rather than me hard code it in as `currentFormValues.run_type = _`. I believe it is a cleaner way of doing it

src/utils/index.js
mikix: I dunno. I mean, on one hand, it'll probably do the right thing most of the time (where the user is just clicking around, hasn't gotten invested in the course). On the other hand, if someone is switching from like a professional course type they've been planning for a month to a master's course type - is there a sane default in that case? What would we expect them to choose? When would that even realistically happen?

I'm guessing we can leave it like you've got it for now and see how it goes. Worst case is someone has to make a new run with the right type, if they go all the way through a review before noticing.

Dillon-Dumesnil: Well hopefully the PC will catch it if it is incorrect and they can go back to unpublished to cause the change. I can also ping Julia real quick about this question since it might be more open than we want.

Dillon-Dumesnil: I'm going to update to not give a default on the edit page so if they change there, they are forced to choose the run type since it's the only time it is not on the page and visible by default

PR #338, opened 2019-11-08
DISCO-1439: Use Type Whitelist in FAP

Dillon-Dumesnil:


src/components/CreateCoursePage/index.jsx
```diff
@@ -10,6 +10,7 @@ import ConfirmationModal from '../ConfirmationModal';
 
 import { formatPriceData } from '../../utils';
 
+const typeWhiteList = process.env.TYPE_WHITELIST ? process.env.TYPE_WHITELIST.split(',') : [];
```
mikix: A toplevel variable? Alright.

src/components/CreateCoursePage/index.jsx
```diff
@@ -39,6 +40,15 @@ class CreateCoursePage extends React.Component {
     this.setState({ startedFetching: true });
   }
 
+  isOrgWhitelisted() {
+    const userOrgs = this.props.publisherUserInfo.organizations;
+    if (!typeWhiteList || (typeWhiteList && typeWhiteList.length === 0)) {
```
mikix: `typeWhiteList` will always exist and be truthy, right? The definition at the top guarantees it. So you might only need the length check. And then, you don't even need the length check, right, because the `some` query will gracefully return false if the list never includes the key.

src/components/CreateCoursePage/index.jsx
Dillon-Dumesnil: Otherwise this would get calculated much more often and this is just less calls. Basically copying the way the other whitelist was created.

src/components/CreateCoursePage/index.jsx
```diff
@@ -39,6 +40,15 @@ class CreateCoursePage extends React.Component {
     this.setState({ startedFetching: true });
   }
 
+  isOrgWhitelisted() {
+    const userOrgs = this.props.publisherUserInfo.organizations;
+    if (typeWhiteList.length > 0 && typeWhiteList[0] === '-') {
```
Dillon-Dumesnil: Supports White list by default and Black list if it is prepended by `'-,'`

src/components/CreateCoursePage/index.jsx
```diff
@@ -39,6 +40,15 @@ class CreateCoursePage extends React.Component {
...
+      const typeBlackList = typeWhiteList.slice(1);
+      return userOrgs.some(org => !typeBlackList.includes(org.key));
```
mikix: This means that a multi-org person would be allowed to use types if any of their orgs are not blacklisted. Maybe it should be if like `!userOrgs.any(org => typeBlackList.includes`? I dunno, multi-org people are a weird case.

Dillon-Dumesnil: Yeah, I was thinking about that. I don't know what the right way of doing it is. I think the way I have it is the most correct because it at least mirrors the behavior of the whitelist. Like right now with the whitelist, as soon as one of your orgs is in the whitelist you can see it. Or I update both to `every`?

PR #339, opened 2019-11-09
Fix up switch to org blacklist

jmyatt:


src/components/CourseTable/CourseTable.test.jsx
```diff
@@ -9,9 +9,28 @@ describe('CourseTable', () => {
     expect(component).toMatchSnapshot();
   });
 
-  it('hides table and button when not whitelisted', () => {
-    process.env.ORG_WHITELIST = ['fake'];
-    const component = shallow(<CourseTable />);
+  it('hides table and button when blacklisted', () => {
+    const publisherUserInfo = { organizations: [{ key: 'fake1', name: 'fake_name1' }] };
+    const component = shallow(<CourseTable publisherUserInfo={publisherUserInfo} />);
+    expect(component).toMatchSnapshot();
```
Dillon-Dumesnil: Just want to confirm that the snapshot for this actually does not have the table showing.

That was an issue I was having in my PR, but it looks like you're doing it correctly with the setupTest.js file below.

jmyatt: yes, that snapshot does not have the table 👍

src/components/CourseTable/CourseTable.test.jsx
```diff
@@ -9,9 +9,28 @@ describe('CourseTable', () => {
...
+  });
+
+  it('displays table and button when not blacklisted', () => {
+    const publisherUserInfo = { organizations: [{ key: 'fake2', name: 'fake_name2' }] };
+    const component = shallow(<CourseTable publisherUserInfo={publisherUserInfo} />);
+    expect(component).toMatchSnapshot();
+  });
+
+  it('displays table and button when no blacklist exists', () => {
+    process.env.ORG_BLACKLIST = [];
```
Dillon-Dumesnil: I don't think this actually overwrites the value since it is set upon import. So this test probably isn't doing what we expect it to.

jmyatt: ah ok thanks - I'll take a look

PR #341, opened 2019-11-12
Fix for read property '1' of undefined

Dillon-Dumesnil:


src/components/CreateCourseRunPage/index.jsx
```diff
@@ -138,7 +138,8 @@ class CreateCourseRunPage extends React.Component {
     const parsedTypeOptions = courseOptionsData && courseOptionsData.type &&
       parseCourseTypeOptions(courseOptionsData.type.type_options);
     const courseRunTypeOptions = parsedTypeOptions && parsedTypeOptions.courseRunTypeOptions;
-    const defaultRunType = courseRunTypeOptions && courseRunTypeOptions[type][1].value;
+    const defaultRunType = courseRunTypeOptions && courseRunTypeOptions[type] &&
+      courseRunTypeOptions[type][1].value;
```
mikix: Do you also want to add a `&& courseRunTypeOptions[type][1]`?

Dillon-Dumesnil: Can't hurt. Going to update test too

src/components/CreateCourseRunPage/index.jsx
```diff
@@ -138,7 +138,8 @@ class CreateCourseRunPage extends React.Component {
...
```
xnick421: defaultRunType will be `undefined` if there are not atleast 2 elements in the `courseRunTypeOptions[type]` array.  is this expected?

Dillon-Dumesnil: Yeah, no problems with that. Just sets the initialValues array and it being undefined there will just leave it as forcing them to select an option.

xnick421: great

PR #349, opened 2019-11-19
Fix infinite spinning submit button

Dillon-Dumesnil:
The problem was we would set a boolean that we are submitting
a run before waiting for html required fields to be caught.
This change makes it so we do not set that boolean to true until
we are actually submitting the course run and not just when the
button is clicked


src/components/EditCoursePage/index.jsx
```diff
@@ -448,6 +450,7 @@ class EditCoursePage extends React.Component {
       !IN_REVIEW_STATUS.includes(targetRun.status)) {
       // Submitting Run for review, show modal, and temporarily store form data until
       // we have a response for how to continue.
+      store.dispatch(courseRunSubmitting());
```
xnick421: I honestly agree with this, but the current pattern right now is to pass in actions as props from the container.  idk i honestly think this is so much easier to read, but its not done like this in our project very often

PR #350, opened 2019-11-20
Only require the OFAC field if that run is being reviewed

Dillon-Dumesnil:


src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -640,7 +640,7 @@ class CollapsibleCourseRun extends React.Component {
                   />
                 }
                 extraInput={{ onInvalid: this.openCollapsible }}
-                required
+                required={jsonDeepEqual(courseSubmitInfo.targetRun, courseRun)}
```
Dillon-Dumesnil: Thoughts on this vs `courseSubmitInfo.targetRun && courseSubmitInfo.targetRun.key === courseRun.key`?

jlajoie: I think what you already have is preferred, if something else changes they wouldn't be the "same" (even though that shouldn't happen).

PR #353, opened 2019-11-21
Hide fields in non-marketable courses

mikix:
If the course is currently in a non-marketable type (no run types have is_marketable=True), then hide a bunch of course-level marketing info like descriptions and outcomes.

https://openedx.atlassian.net/browse/DISCO-1462


src/components/EditCoursePage/EditCourseForm.test.jsx
```diff
@@ -189,8 +189,8 @@ describe('BaseEditCourseForm', () => {
       id="edit-course-form"
     />);
 
-    const disabledFields = component.find({ disabled: true });
-    expect(disabledFields.length === 1);
+    const disabledFields = component.find({ name: 'mode', disabled: true });
+    expect(disabledFields).toHaveLength(1);
```
mikix: `expect(disabledFields.length === 1)` will never fail. Had to fix some tests once I noticed they were using this pattern.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -390,380 +394,382 @@ export class BaseEditCourseForm extends React.Component {
               className="course-image"
               disabled={disabled}
             />
-            <hr />
```
mikix: The rest of this file is just wrapping a huge section in `{showMarketingFields && <React.Fragment>` and indenting.

src/components/EditCoursePage/EditCourseForm.test.jsx
```diff
@@ -254,7 +254,41 @@ describe('BaseEditCourseForm', () => {
...
-    const disabledFields = component.find({ disabled: true });
-    expect(disabledFields.length === 1);
+    const disabledFields = component.find({ name: 'type', disabled: true });
+    expect(disabledFields).toHaveLength(1);
+  });
+
+  it('no marketing fields if course type is not marketable', () => {
+    const initialValuesWithMasters = {
+      ...initialValuesFull,
+      type: '7b41992e-f268-4331-8ba9-72acb0880454',
```
jlajoie: question/tiniest of nits: Is type supposed to be a UUID here, not a whole type object? Maybe call it type_uuid?

jlajoie: Side note, for all of our testData.js, should these be better as something more like aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee? I assume these are just UUIDs we pulled from our local stacks to use as static test data but just wanted to be sure!

mikix: > Is type supposed to be a UUID here, not a whole type object? Maybe call it type_uuid?

Yeah, it's a UUID that refers to a full CourseType object on the backend. I'm not interested in renaming it now in this PR, but fair question. :)

> Side note, for all of our testData.js, should these be better as something more like aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee?

That's what I would have done if manually entering these. But this data started with a copy/paste from a devstack, and it was easier to leave uuids as is, I suspect.

One bummer of UUIDs is that it's less readable for sure. Like this would like nicer if it was `type: 'masters-verified-audit'` or some such.

PR #354, opened 2019-11-22
Filter on unsubmitted, not unpublished

mikix:
https://openedx.atlassian.net/browse/DISCO-1455

Tied to backend PR https://github.com/edx/course-discovery/pull/2302


mikix:
> Seems reasonable, but should we add 'unsubmitted' to our constants as well just for consistency?

:man_shrugging: We didn't for `in_review` so I was just aping that. I guess one argument is that it's not a status used elswhere, it's only accepted by this one endpoint. So maybe keeping it as not a constant prevents spread of it? But a comment could probably address that issue. I dunno, I'm lazy and going to leave it for now.

PR #371, opened 2019-12-13
Drop non-course-type paths

mikix:
https://openedx.atlassian.net/browse/DISCO-1399


src/utils/index.js
```diff
@@ -110,6 +110,20 @@ const getErrorMessages = (error) => {
   return ['Unknown error.'];
 };
 
+const pricesEqual = (leftPrices, rightPrices) => {
```
Dillon-Dumesnil: Can we rename these arguments to `initialPrices` and `currentPrices`? I don't love left and right. Same goes for all the left and right throughout the function

Dillon-Dumesnil:
Also FYI didn't review snapshots. Assuming you're still working through them given the failing tests

src/utils/index.js
mikix: Hrm. I wrote this in the mindset that it was a generic compare method. Like if I wrote a `strcmp` function, I wouldn't name the arguments after one particular use of it.

But since this is so tightly bound to its usage in this file (and isn't even exported), I'm down with renaming to make it more readable.

Dillon-Dumesnil: I'm good with keeping it general. Left right feels weird. Maybe old new? I don't know. I trust your judgement

PR #373, opened 2019-12-19
Display times in local timezone

mikix:
We were displaying times as UTC, even though we told the user it was in a local timezone. This fixes that.

https://openedx.atlassian.net/browse/DISCO-1518

### DIFF
--- a/src/components/EditCoursePage/CollapsibleCourseRun.jsx
+++ b/src/components/EditCoursePage/CollapsibleCourseRun.jsx
@@ -219,7 +219,7 @@ class CollapsibleCourseRun extends React.Component {
   normalizeDate(date) {
     if (date) {
       if (date.match(NORMALIZE_DATE_MATCHER)) {
-        return moment.utc(date).format(DATE_FORMAT);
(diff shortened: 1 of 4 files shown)


src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -216,10 +217,16 @@ class CollapsibleCourseRun extends React.Component {
     return '';
   }
 
+  dayToIso(day) {
+    // 'date' is a day-only string (or UTC ISO formatted string).
+    // We want to end up with an ISO formatted date that is at 0:00 UTC on that same day.
+    return moment.utc(day).format(DATE_FORMAT);
```
mikix: This is same code as before, I just busted it out to a function so I can add a comment in one place.

Basically, we want to keep treating go_live_date as a UTC date, because that field isn't really a datetime, it's just a day. And we say "we'll publish this run *sometime* during that day". So it's not really timezone relevant. And while we want to give the backend midnight UTC timestamps for this field, we don't want to accidentally render midnight UTC as a different day based on timezone.

PR #377, opened 2020-01-16
DISCO-1551 add plugin to allow users to specify text language

karengu:
ezgif com-video-to-gif (1)

uses plugin from https://github.com/edx/tinymce-language-selector/pull/1


src/components/RichEditor/LanguagePlugin.js
```diff
@@ -0,0 +1,132 @@
...
+        }
+        store.dispatch(updateFormLanguage(data.language, editor.id));
+        buttonApi.setActive(data.language !== BROWSER_DEFAULT);
+        api.close();
+      },
+    });
+  };
+
+  editor.ui.registry.addToggleButton('language', {
+    text: 'Foreign language',
```
Dillon-Dumesnil: Still think we need to figure out a way to dynamically change this otherwise it might not be a great UX

src/components/RichEditor/LanguagePlugin.js
```diff
@@ -0,0 +1,132 @@
...
+        const state = store.getState();
+        if (getLanguageforEditor(state, editor.id) !== BROWSER_DEFAULT) {
+          resetLanguage();
+          buttonApi.setActive(false);
+          store.dispatch(updateFormLanguage(BROWSER_DEFAULT, editor.id));
+        }
+      });
+      const updateCurrentLanguage = () => {
+        const selectedNode = editor.selection.getNode();
+        if (selectedNode.lang) {
```
Dillon-Dumesnil: This could be improved through a ternary expression.

src/components/RichEditor/LanguagePlugin.js
```diff
@@ -0,0 +1,132 @@
+import tinymce from 'tinymce/tinymce.min';
+import { BROWSER_DEFAULT, languages } from '../../data/constants/formOptions';
+import updateFormLanguage from '../../data/actions/formOptions';
+import { getLanguageforEditor } from '../../data/reducers/formOptions';
+import store from '../../data/store';
+
+tinymce.PluginManager.add('language', (editor) => {
```
Dillon-Dumesnil: I think it would be helpful to leave comments throughout this file explaining what is happening at different locations. Places I think that could use commenting:

1. All of the `selectedNode.nodeName` if conditions and the differences in the results there.
2. the `resetLanguage` function

If there are any other spots you think could be unclear, please also comment those.

src/components/RichEditor/LanguagePlugin.js
```diff
@@ -0,0 +1,132 @@
...
+    });
+  };
+
+  editor.ui.registry.addToggleButton('language', {
+    text: 'Foreign language',
+    onAction(buttonApi) {
+      openDialog(buttonApi);
+    },
+    onSetup(buttonApi) {
+      editor.addShortcut('Meta+L', 'Switch to default language', () => {
```
Dillon-Dumesnil: Nit: I think it might be better just not including this... It feels like potentially weird behavior that Meta+L will take you out, but hitting it again won't put you back in. I think it is correct to not assume their language and not allowing them to Meta+L into a foreign language, but it might be better to just not include at all. Would like to hear your thoughts on this and potentially we can check in with others as well.

karengu: I could see this either way; I figured if people are switching in and out of a language often, they would find it annoying to have to open up the menu to switch back to the default.

src/components/RichEditor/LanguagePlugin.js
karengu: Might be able to do something like this:
http://archive.tinymce.com/forum/viewtopic.php?id=33299
but the fact that we have multiple editors complicates things.

PR #378, opened 2020-01-24
Add a special 504 message on reruns

mikix:
As a stop-gap measure, we give a special message when receiving a 504 message back from Discovery when making a rerun.

I didn't add a test, because we still don't seem to have a good way to test network interactions. Seems bad, yah?

https://openedx.atlassian.net/browse/DISCO-1548

Screenshot from 2020-01-24 15-00-02

### DIFF
--- a/src/data/actions/courseInfo.js
+++ b/src/data/actions/courseInfo.js
@@ -154,7 +154,15 @@ function createCourseRun(courseUuid, courseRunData) {
         return dispatch(push(`/courses/${courseUuid}`));
       })
       .catch((error) => {
-        dispatch(createCourseRunFail(['Course Run create failed, please try again or contact support.'].concat(getErrorMessages(error))));
+        let errorList;
+        if (error.response && error.response.status === 504) {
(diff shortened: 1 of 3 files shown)


src/data/actions/courseInfo.js
```diff
@@ -154,7 +154,15 @@ function createCourseRun(courseUuid, courseRunData) {
         return dispatch(push(`/courses/${courseUuid}`));
       })
       .catch((error) => {
-        dispatch(createCourseRunFail(['Course Run create failed, please try again or contact support.'].concat(getErrorMessages(error))));
+        let errorList;
+        if (error.response.status === 504) {
```
jlajoie: Might be worth confirming there's a response object here?

jlajoie:
As an aside, should we maybe up the margin in that image between the error box and the buttons? They seem a little snug.

src/data/actions/courseInfo.js
mikix: Done.

mikix:
> As an aside, should we maybe up the margin in that image between the error box and the buttons? They seem a little snug.

Good point, done.

PR #385, opened 2020-02-06
WS-624 make url slug field read-only for non-staff users in publisher

rgraber:


src/components/CreateCoursePage/__snapshots__/CreateCourseForm.test.jsx.snap
```diff
@@ -645,7 +645,7 @@ ShallowWrapper {
               pattern="20[1-9][0-9]/(0[1-9]|1[012])/(0[1-9]|[12][0-9]|3[01])"
               placeholder="yyyy/mm/dd"
               required={true}
-              timeLabel="Start time (UTC)"
```
albemarle: I believe UTC is actually correct here?

albemarle: Or is it just that PCs think of it as GMT?

PR #394, opened 2020-02-21
Support frontend-platform

mikix:
Switch to latest version of frontend-platform and footer rather than the bundle of independent (and deprecated) individual libraries that have all been swept up into the platform repo.

https://openedx.atlassian.net/browse/DISCO-1454


config/webpack.dev.config.js
```diff
@@ -109,23 +109,6 @@ module.exports = Merge.smart(commonConfig, {
       ACCESS_TOKEN_COOKIE_NAME: 'edx-jwt-cookie-header-payload',
       USER_INFO_COOKIE_NAME: 'edx-user-info',
       SITE_NAME: 'edX',
-      MARKETING_SITE_BASE_URL: 'http://localhost:18000',
```
mikix: Dropped these env vars because the footer no longer needs them - it hardcodes this stuff.

package.json
```diff
@@ -22,18 +22,16 @@
   },
   "license": "AGPL-3.0",
   "dependencies": {
-    "@edx/frontend-auth": "5.3.6",
-    "@edx/frontend-component-footer": "4.1.5",
-    "@edx/frontend-logging": "3.0.1",
+    "@edx/frontend-component-footer-edx": "3.0.8",
```
mikix: On the advice of fedx, we're hardcoding the edx version for now (we currently hardcode the edx CSS theme already). This is because fedx does not like the current way of switching between the two footers for MFEs (which involves build time swapping out of npm dependencies, so it's hard to test with the final product).

When we make this repo more open edx friendly, we'll have to revisit the question of theming.

src/data/services/DiscoveryDataApiService.js
```diff
@@ -12,7 +13,7 @@ class DiscoveryDataApiService {
       exclude_utm: 1,
     };
     const url = `${discoveryBaseUrl}/courses/${uuid}/`;
-    return apiClient.get(url, {
+    return getAuthenticatedHttpClient().get(url, {
```
Dillon-Dumesnil: Should this file just instantiate this once at the top rather than per call? Or is there a potential token issue or something that way?

mikix: It does some kind of async lookup - such that I think it's good to not do it globally anymore? Like, it's only defined after initialization time? And I don't know when globals get defined. I saw another frontend do this pattern, so I just stole it.

PR #396, opened 2020-02-21
Use frontend-build

mikix:


webpack.dev.config.js
```diff
@@ -0,0 +1,7 @@
+const { createConfig } = require('@edx/frontend-build');
+
+const config = createConfig('webpack-dev');
+
+config.module.rules[0].exclude = /node_modules\/(?!(tinymce-language-selector|@edx))/;
```
Dillon-Dumesnil: Do these not need to be in a string?

mikix: Naw, that's like a javascript regex native?

PR #442, opened 2020-04-14
Re-publish runs by default on Save

mikix:
That is, when the Save button is pressed, if there are changes to an existing published run, we'll republish it automatically.

This also removes the button to specifically republish a single run.

Hopefully this will avoid users queuing up draft changes and not realizing they aren't published.

https://openedx.atlassian.net/browse/DISCO-1555

### Published run, no edits yet
Screenshot from 2020-04-16 11-13-38

### Published run with edits (at the course or run level)
Screenshot from 2020-04-16 11-13-50

### Unpublished run with edits (same as before)
Screenshot from 2020-04-16 11-14-46


jmyatt:
Let's hold off on merging this for the moment while we confirm the Product perspective

PR #443, opened 2020-04-16
Drop redundant dependencies

mikix:
Some dependencies we are already getting from frontend-build or frontend-platform and don't need to re-specify. This will reduce our renovate churn.


package.json
```diff
@@ -25,14 +25,12 @@
...
     "@edx/paragon": "7.2.1",
-    "@fortawesome/fontawesome-svg-core": "1.2.28",
     "@fortawesome/free-solid-svg-icons": "5.13.0",
     "@fortawesome/react-fontawesome": "0.1.9",
     "@tinymce/tinymce-react": "3.5.0",
     "classnames": "2.2.6",
-    "core-js": "3.6.5",
     "connected-react-router": "6.8.0",
-    "email-prop-type": "3.0.1",
+    "core-js": "3.6.5",
```
Dillon-Dumesnil, line 33: I don't trust this change.

Dillon-Dumesnil: Because it is the exact same as the line that was removed (couple lines up)

PR #507, opened 2020-09-08
WS-1293: Gracefully handle Collaborators who are missing names or images.

reillz10:
… url is provided.

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -190,7 +190,7 @@ const getOptionsData = (options) => {
 };
 
 const formatCollaboratorOptions = (options) => (
-  options.map(({ name, uuid, image: { original: { url } } }) => ({ name, uuid, image_url: url }))
+  options.map(({ name = '', uuid, image: { original: { url } = { url: '' } } }) => ({ name, uuid, image_url: url }))
 );
 
 const parseCourseTypeOptions = (typeOptions) => {


rgraber (approved):
Would dearly like to know how these are getting deleted, but this makes sense to me

Dillon-Dumesnil:
Should the collaborator field just be not required?

rgraber:
> Should the collaborator field just be not required?

It doesn't make any sense to have a collaborator without an image url, since the whole point of a collaborator is to add logos to a section of the course about page. I think we should leave it required, and ideally at some point figure out how they are getting deleted and prevent it

reillz10:
> Should the collaborator field just be not required?

@Dillon-Dumesnil I don't think that it is required. The problem is with the list of collaborators that get passed to publisher as options to be added onto the course. If one of those is missing a url then the application blows up. This is to avoid the application blowing up.

src/utils/index.js
```diff
@@ -190,7 +190,7 @@ const getOptionsData = (options) => {
 };
 
 const formatCollaboratorOptions = (options) => (
-  options.map(({ name, uuid, image: { original: { url } } }) => ({ name, uuid, image_url: url }))
+  options.map(({ name, uuid, image: { original: { url = 'https://www.edx.org/images/edx-logo-header.png' } } }) => ({ name, uuid, image_url: url }))
```
albemarle, line 193: Hmm, thinking about this... is this file the right size or close to it?  Also, it might be a good idea to add a comment as to why this is necessary.  Or maybe there is more debugging we could add to help track this issue down?

reillz10: I can look closer and see how the logos are getting deleted. I'll timebox that effort to mid day tomorrow and see what I can find.

Dillon-Dumesnil:
Okay, I see what you're both saying. I was thinking this was going to be a required field and not simply the image that is associated with a collaborator. Understood why it is required and agree that if you are adding a collaborator, an image should continue to be required.

I will add some push back on setting the edX logo as the default image. I know y'all are investigating the underlying issue, but I feel like using the edX logo is also going to be misleading for our course teams when they come back in to view their Publisher course (or even more important, their marketing page) and see the desired text for a collaborator with an edX logo over it. 

In the interim, perhaps it is better that if the collaborator data is bad (image has mysteriously disappeared), we don't show that collaborator or an error message or something similar. Simply placing edX over does not feel like the correct solution to me.

reillz10:
@rgraber @Dillon-Dumesnil Would you mind taking a look at this again? I have updated the PR to set the url and name to an empty string if the api doesn't return one. I'm pretty sure that I have fixed the underlying issue in disco, but I want to make sure that Publisher doesn't completely go down if I'm wrong about that fix.

rgraber:
> @rgraber @Dillon-Dumesnil Would you mind taking a look at this again? I have updated the PR to set the url and name to an empty string if the api doesn't return one. I'm pretty sure that I have fixed the underlying issue in disco, but I want to make sure that Publisher doesn't completely go down if I'm wrong about that fix.

What does this end up looking like in the UI?

Dillon-Dumesnil:
> @rgraber @Dillon-Dumesnil Would you mind taking a look at this again? I have updated the PR to set the url and name to an empty string if the api doesn't return one. I'm pretty sure that I have fixed the underlying issue in disco, but I want to make sure that Publisher doesn't completely go down if I'm wrong about that fix.

If the API doesn't return one, do you need to even set it to an empty string? Won't it just be `null`? What's the difference in behavior?

reillz10:
> > @rgraber @Dillon-Dumesnil Would you mind taking a look at this again? I have updated the PR to set the url and name to an empty string if the api doesn't return one. I'm pretty sure that I have fixed the underlying issue in disco, but I want to make sure that Publisher doesn't completely go down if I'm wrong about that fix.
> 
> If the API doesn't return one, do you need to even set it to an empty string? Won't it just be `null`? What's the difference in behavior?

Currently, publisher goes down on load for all courses when there is a missing url with `Cannot read property ... of undefined`. This stops all PCs and course teams from doing things in publisher even if they might not be doing anything with Collaborators. This is in place to not have the whole application fall over and we can handle errors in our own time, rather than having it be a fire drill when it does go down. The way this works currently, when an image url is missing then we will just see the name of the collaborator and vice versa. Whomever is adding the collaborator (PC's and marketing are the ones using this now) will likely notice this issue and we can help them in our own time, rather than having to drop everything to take a look.

rgraber:
> > > @rgraber @Dillon-Dumesnil Would you mind taking a look at this again? I have updated the PR to set the url and name to an empty string if the api doesn't return one. I'm pretty sure that I have fixed the underlying issue in disco, but I want to make sure that Publisher doesn't completely go down if I'm wrong about that fix.
> > 
> > 
> > If the API doesn't return one, do you need to even set it to an empty string? Won't it just be `null`? What's the difference in behavior?
> 
> Currently, publisher goes down on load for all courses when there is a missing url with `Cannot read property ... of undefined`. This stops all PCs and course teams from doing things in publisher even if they might not be doing anything with Collaborators. This is in place to not have the whole application fall over and we can handle errors in our own time, rather than having it be a fire drill when it does go down. The way this works currently, when an image url is missing then we will just see the name of the collaborator and vice versa. Whomever is adding the collaborator (PC's and marketing are the ones using this now) will likely notice this issue and we can help them in our own time, rather than having to drop everything to take a look.

This makes sense to me but I'd still like a screenshot from the UI when something returns empty, just to make sure it doesn't look absolutely bonkers

Dillon-Dumesnil:
> > > @rgraber @Dillon-Dumesnil Would you mind taking a look at this again? I have updated the PR to set the url and name to an empty string if the api doesn't return one. I'm pretty sure that I have fixed the underlying issue in disco, but I want to make sure that Publisher doesn't completely go down if I'm wrong about that fix.
> > 
> > 
> > If the API doesn't return one, do you need to even set it to an empty string? Won't it just be `null`? What's the difference in behavior?
> 
> Currently, publisher goes down on load for all courses when there is a missing url with `Cannot read property ... of undefined`. This stops all PCs and course teams from doing things in publisher even if they might not be doing anything with Collaborators. This is in place to not have the whole application fall over and we can handle errors in our own time, rather than having it be a fire drill when it does go down. The way this works currently, when an image url is missing then we will just see the name of the collaborator and vice versa. Whomever is adding the collaborator (PC's and marketing are the ones using this now) will likely notice this issue and we can help them in our own time, rather than having to drop everything to take a look.

So that makes it sound like the issue is that you're passing in undefined results to `formatCollaboratorOptions`. I suppose this is one way around that since even when there is no options defined, you're setting default values. To help avoid that sort of error in the future, you can also check and make sure the data exists prior to calling the format function.  But that might just be a matter of style.

Also interesting it crashes all of Publisher for everyone on something missing. I would have thought it would only affect the one page that it is failing on. Something to think on for another time I suppose.

Agreed with Becca on having a screenshot, but otherwise this PR seems fine

reillz10:
@rgraber just a sec, my devstack is acting up and I cant log in again. I'll get you some ss in a few minutes.

reillz10:
@rgraber @Dillon-Dumesnil 
When there is a logo (the logo i'm using is a screen shot of my iterm terminal):
Screen Shot 2020-09-10 at 2 11 43 PM


When there isn't a logo: 
image

rgraber (approved):
LGTM

Dillon-Dumesnil:
@reillz10 And is the intent here that you would never actually save with that value? Because that would then send back the empty string as the value for the collaborator. Is there any frontend validation to prevent that or relying on the backend to catch it?

reillz10:
@Dillon-Dumesnil yes, we would never actually save with that value. Again, I think that the root cause of the missing collaborator logos has been fixed on the discovery end. There is already frontend validation that you have to have an image in order to submit a collaborator. Somehow, that validation wasn't actually protecting discovery from bad requests. I merged a fix in discovery to actually do server side validations on Post and Patch calls which should have fixed the issue. This is really just a fallback.

Dillon-Dumesnil:
> @Dillon-Dumesnil yes, we would never actually save with that value. Again, I think that the root cause of the missing collaborator logos has been fixed on the discovery end. There is already frontend validation that you have to have an image in order to submit a collaborator. Somehow, that validation wasn't actually protecting discovery from bad requests. I merged a fix in discovery to actually do server side validations on Post and Patch calls which should have fixed the issue. This is really just a fallback. I can jump on a call to show you through this if you want.

👍

Dillon-Dumesnil (approved):

PR #568, opened 2021-04-05
use edx logo from remote config

uzairr:
Use edx logo from remote config instead of hard-code.

VAN-432


src/components/Header/index.jsx
```diff
@@ -6,8 +6,6 @@ import { redirectToLogout } from '@edx/frontend-platform/auth';
 import { AppContext } from '@edx/frontend-platform/react';
 import { Dropdown, DropdownButton, Hyperlink } from '@edx/paragon';
 
-import EdxLogo from '../../../assets/edx-rebrand-logo.png';
```
waheedahmed, line 9: Could we also remove the asset file?

waheedahmed (commented):
You might also need to update the footer version to the latest.

waheedahmed (approved):

PR #570, opened 2021-04-07
update imports from brand package

uzairr:
Add brand package and update its relevant imports.

VAN-233

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "@edx/brand": "npm:@edx/brand-openedx@1.1.0",
     "@edx/frontend-component-footer-edx": "3.1.0",
     "@edx/frontend-platform": "1.9.0",
-    "@edx/paragon": "11.1.0",
+    "@edx/paragon": "13.17.5",
     "@edx/tinymce-language-selector": "1.1.0",
     "@fortawesome/free-solid-svg-icons": "5.15.1",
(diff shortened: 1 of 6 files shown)


waheedahmed (commented):
You need to update the paragon version to the latest alongside these changes otherwise new theme changes won't work.

package.json
```diff
@@ -26,7 +26,7 @@
     "@edx/brand": "npm:@edx/brand-openedx@1.1.0",
     "@edx/frontend-component-footer-edx": "3.0.10",
     "@edx/frontend-platform": "1.9.0",
-    "@edx/paragon": "11.1.0",
+    "@edx/paragon": "^14.4.0",
```
waheedahmed, line 29: Since some components are deprecated in the latest version of paragon, let's not update it to the latest version right now and instead use `13.17.5`. Also, remove the caret sign.

With this version, you might not need to update the test snapshots as well.

waheedahmed: Also, this version is not yet supported by `"@edx/frontend-platform": "1.9.0"`

waheedahmed (changes requested):

package.json
uzairr: thanks @waheedahmed for the context

waheedahmed (approved):

PR #576, opened 2021-04-15
ENT-4435: Display skill tags in Course Edit Page in Publisher

sameenfatima78:
This PR displays skill tags when in course edit page. 

**JIRA:** 

- https://openedx.atlassian.net/browse/ENT-4332
- https://openedx.atlassian.net/browse/ENT-4435

Backend PR: https://github.com/edx/course-discovery/pull/3010

**Example:** 

Screenshot 2021-04-15 at 2 18 10 PM

**Tooltip:**

Screenshot 2021-04-15 at 2 20 08 PM


src/components/EditCoursePage/CourseSkills.jsx
```diff
@@ -0,0 +1,52 @@
...
+        ))}
+      </div>
+    );
+  }
+}
+
+CourseSkills.propTypes = {
+  className: PropTypes.string.isRequired,
+  id: PropTypes.string.isRequired,
+  input: PropTypes.shape({
```
saleem-latif, line 45: shouldn't `input` be renamed to `skills` ?

src/components/EditCoursePage/CourseSkills.jsx
```diff
@@ -0,0 +1,52 @@
...
+    } = this.props;
+    const {
+      skillNames,
+    } = this.state;
+    return (
+      <div className="form-group mb-2">
+        <div id={id} name={name} tabIndex="-1">{label}</div>
+        {skillNames.map(skill => (
+          <Badge
+            key={skill.id}
```
saleem-latif, line 30: `skillNames` is a list of strings so `skill.id` will return null. If skill is guaranteed to be unique which I think is, you can use `key={skill}` here.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -875,6 +881,32 @@ export class BaseEditCourseForm extends React.Component {
...
+                        edX partners with Emsi, the labor market data company, to automatically tag your courses with
+                        in-demand skills from their library of 30,000 skills based on the content in your about page.
+                        If you want to experiment with what skills show up, you can edit and submit changes to your
+                        about page description.
+                      </p>
+                    </div>
+                  )}
+                />
+              )}
+              disabled={disabled}
```
Dillon-Dumesnil, line 905: Shouldn't disabled always be true? It looked like a read only field.

Dillon-Dumesnil (commented):
This all looks fine to me. Had a small question about the disabled field. Also wondering about how the boolean field `enable_skills_in_publisher` is going to end up being used. I don't see it being put into effect here or in course discovery at the moment.

src/components/EditCoursePage/EditCourseForm.jsx
sameenfatima78: Good catch, I have fixed it.

src/components/EditCoursePage/CourseSkills.jsx
sameenfatima78: This built in prop is directly coming from the parent component. Every field inside a `redux-form` has an `input` prop coming in.

src/components/EditCoursePage/CourseSkills.jsx
sameenfatima78: Done

sameenfatima78:
@Dillon-Dumesnil Sorry this was a work in progress PR and I couldn't manage to wrap up the changes on that same day. I have pushed new changes now with the feature flag `enable_skills_in_publisher` in use and incorporated your feedback. Please can you review again?

zamanafzal (approved):
LGTM 👍🏻

saleem-latif (approved):
Looks Good 👍

PR #586, opened 2021-06-25
ENT-4568: skills are defaulted visible to all partners

sameenfatima78:
Removed logic that decides whether or not skills show up based on the flag `enable_skills_in_publisher`. Should be defaulted visible to all partners now. 

JIRA: https://openedx.atlassian.net/browse/ENT-4568


src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -884,7 +881,7 @@ export class BaseEditCourseForm extends React.Component {
               disabled={disabled}
               optional
             />
-            {displaySkills
+            {skillNames?.length > 0
```
saleem-latif, line 884: do we want to add `skillNames` not null check here i.e. `skillNames && skillNames.length > 0` is that is not necessary here?

saleem-latif (approved):
One question, other than that looks good 👍🏼

src/components/EditCoursePage/EditCourseForm.jsx
sameenfatima78: The [Optional Chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) operator (?) will take care of that.

PR #588, opened 2021-09-10
fix: add upgrade deadline override field

mubbsharanwar:
- add upgrade deadline date field in course run form.
- Only staff user can edit this field.

### Form changes:
Screen Shot 2021-09-10 at 7 54 00 PM

### Other nots:
This PR is dependent on 
https://github.com/edx/course-discovery/pull/3123/

DISCO-1622

### DIFF
--- a/src/components/EditCoursePage/EditCoursePage.test.jsx
+++ b/src/components/EditCoursePage/EditCoursePage.test.jsx
@@ -23,6 +23,7 @@ const mockStore = configureStore();
 describe('EditCoursePage', () => {
   const defaultPrice = '77';
   const defaultEnd = '2019-08-14T00:00:00Z';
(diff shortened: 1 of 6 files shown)


waheedahmed (changes requested):
The field should be `upgrade_deadline_override` instead of `upgrade_deadline`, otherwise looks good.

src/components/EditCoursePage/EditCoursePage.test.jsx
```diff
@@ -633,14 +642,14 @@ describe('EditCoursePage', () => {
       component.instance().getData = jest.fn();
       component.update();
 
-      courseData.prices.verified = '500.00';
-      expectedSendCourse.prices.verified = '500';
```
waheedahmed, line 637: Why did we change these values?

waheedahmed (approved):

PR #662, opened 2022-02-17
feat: make staff optional for exec ed courses

azanbinzahid:
## Description:
This PR will add a check to make `staff` field optional for `Executive Education` courses. 

## Linked Ticket:
[PROD-2685](https://openedx.atlassian.net/browse/PROD-2685)

## Logic:
Validation function only receive track `type`'s uuid (as in disco) under `targetRun` prop therefore I can't seem to find an easier way to determine track type name in validation function. One option is to take `mode` from `entitlement` prop but that only works if correct track is set at creation (e.g. not if we are editing from Audit to Executive Education). Hence, this `prices.<track-slug>` field only have a positive count when desired track (i.e. executive education) is selected. However, if there are better options available, we should opt that or make sure that the underlying assumption is correct before merging this PR. 

## TODO:
- [x] Add unit tests
- [ ] Improve exec ed track check
- [x] Get owner's review
- [ ] The name of `Executive Education` track may be renamed to `Executive Program`

### DIFF
--- a/src/utils/validation.js
+++ b/src/utils/validation.js
@@ -73,7 +73,7 @@ const handleCourseEditFail = (errors) => {
 };
 
 const editCourseValidate = (values, props) => {
-  const { targetRun } = props;
+  const { targetRun, registeredFields } = props;
 
   if (!targetRun || targetRun.status === PUBLISHED) {
     return {};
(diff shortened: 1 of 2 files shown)


DawoudSheraz (approved):
Tested locally, working as expected. Although we can double-check for other devs on the approach taken here, just to be sure if everything is ok.

src/utils/validation.js
```diff
@@ -98,7 +98,12 @@ const editCourseValidate = (values, props) => {
     const { key: targetKey } = targetRun;
     const isSubmittingRun = run.key === targetKey;
     if (isSubmittingRun) {
-      const runRequiredFields = ['transcript_languages', 'staff'];
+      // naive check to deteremine if track is executive education or not
+      const isExecutiveEducation = registeredFields
+        && registeredFields['prices.executive-education']
+        && registeredFields['prices.executive-education'].count;
```
waheedahmed, line 104: I guess the count condition isn’t necessary, it cannot be 0 if the `prices.executive-education` is set.

waheedahmed (approved):
One minor suggestion otherwise 👍

src/utils/validation.js
azanbinzahid: While testing, I have observed that in `Edit Course Page` while we are still editing the form and change course enrollment track from audit to exec ed to audit again, we still see price for exec ed but with 0 count. Not having a count check will fail validation for audit only course as well. 

image

PR #703, opened 2022-04-06
fix: Change upgrade_deadline timezone to UTC

syedsajjadkazmii:
Change upgrade_deadline timezone in publisher from
local timezone to UTC timezone

[DISCO-1803](https://openedx.atlassian.net/browse/DISCO-1803)


shafqatfarhan (approved):
👍  LGTM

PR #723, opened 2022-04-21
feat: add navigation to course groups

azanbinzahid:
## Description:
This PR will add navigation routes to different course filters. 

## Linked Ticket:
[PROD-2780](https://openedx.atlassian.net/browse/PROD-2780)

## Logic:
Existing implementation for query parameters handling in course tables page is little tricky. It caters filters and search side by side and some utility functions are naive e.g. filtering of `selectedFilters` from component state while updating url. In order to work with existing implementation and not break them, this solution has been added this way. This could be improved along with updating logic for handling existing query parameters and filters. 

## TODO:
- [ ] add unit test once the approach is accepted
- [x] highlight navigation route in use
- [ ] use state variable for course type if needed
- [x] resolve issue with changing url when route double clicked (UI + removes existing filters)
- [ ] comment approach if existing handling is updated

### DIFF
--- a/src/components/CourseTable/index.jsx
+++ b/src/components/CourseTable/index.jsx
@@ -59,6 +59,7 @@ class CourseTable extends React.Component {
       },
     } = this.props;
     const { administrator } = getAuthenticatedUser();
+    const { selectedFilters } = this.state;
     const prevEditorFilterOptions = prevProps.table.editorFilterOptions;
 
     if (editorFilterOptions !== prevEditorFilterOptions && !administrator) {
@@ -71,13 +72,18 @@ class CourseTable extends React.Component {
     const {
       editors: prevEditors,
       course_run_statuses: prevCourseRunStatuses,
+      course_type: prevCourseType,
     } = qs.parse(prevProps.location.search);
     const {
       editors,
       course_run_statuses: courseRunStatuses,
+      course_type: courseType,
     } = qs.parse(this.props.location.search);
     if ((editors !== prevEditors) || (courseRunStatuses !== prevCourseRunStatuses)) {
       this.getSelectedFiltersFromUrl();
+    }
+    if (courseType !== prevCourseType) {
+      this.updateFilterQueryParamsInUrl(selectedFilters);
     }
   }
 
--- a/src/components/Header/index.jsx
+++ b/src/components/Header/index.jsx
@@ -19,7 +19,13 @@ const Header = ({ darkModeOn, location, toggleDarkMode }) => {
 
   // Allow users to toggle dark mode on if they put in a querystring like ?bananas=1
   const querystringParams = qs.parse(location.search);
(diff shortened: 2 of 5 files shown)


DawoudSheraz (commented):
I haven't tested it out on the local yet. Missing unit tests.

src/components/Header/index.jsx
```diff
@@ -30,9 +30,16 @@ const Header = ({ darkModeOn, location, toggleDarkMode }) => {
               <img src={process.env.LOGO_URL} alt="edX logo" height="30" width="60" />
             </Hyperlink>
           </div>
-          <div className="col">
-            <Link to="/">Courses</Link>
+          <div className="col-auto justify-content-start">
+            <Link to="/">Open Courses</Link>
+          </div>
+          <div className="col-auto justify-content-start">
+            <Link to="/?course_type=executive-education-2u">Exective Education</Link>
```
DawoudSheraz, line 37: spelling: Executive

DawoudSheraz:
@azanbinzahid I have run the code on local, a few observations:

- The filters are getting reset upon the nav switch, as mentioned in the description. We should keep the filters persisted across nav changes. We can look into the existing filters as to how they are persisted when other filter values are changed.
- The double click filter issue would be resolved once we figure out the filter reset issue.
- The approach is sort of consistent with the current filter mechanism. Revamping the filter process will be a significant effort.
- [ ] To verify if the filter is affecting API call time. I noticed slowness in API response time on front end. It could be my system but worth verifying

src/components/CourseTable/index.jsx
```diff
@@ -71,14 +72,19 @@ class CourseTable extends React.Component {
...
     const {
       editors,
       course_run_statuses: courseRunStatuses,
+      course_type: courseType,
     } = qs.parse(this.props.location.search);
     if ((editors !== prevEditors) || (courseRunStatuses !== prevCourseRunStatuses)) {
       this.getSelectedFiltersFromUrl();
     }
+    if (courseType !== prevCourseType) {
+      this.updateFilterQueryParamsInUrl(selectedFilters);
```
DawoudSheraz, line 86: course editor and run status are coming into CourseTable component, not managed by the table component. If we manage state for type, can't it mitigate the filter reset issue? or am I missing something here?

azanbinzahid:
@DawoudSheraz to the first comment, as per internal communication - nav route has been disabled on first click to avoid double click and persist filters and query params. Response time of api might be due to filtering on backed.  
Secondly: We can add a state variable for `course_type` but that won't add much to existing mechanism of `selectedFilters` unless we rewrite the logic. 

I believe for the scope of this ticket, we are in-line with existing implementation. However this can be improved in upcoming phases of redesign.

AliAdnanSohail (approved):

waheedahmed (approved):

PR #737, opened 2022-06-21
feat: Create a Course Pane for Subscription Toggle 

kiram15:
Add behavior to toggle the Subscription Catalog inclusion boolean. This behavior should only be shown when the user has is_staff = true permissions. This also includes the work to be done with the backend functionality of the metadata updating. This endpoint should ensure on the backend that they are staff for security purposes.

**Screenshot with info hover**
Screen Shot 2022-06-21 at 5 03 15 PM

**Error state**
Screen Shot 2022-06-21 at 5 03 48 PM


kiram15:
Closed for forking issues with SOX changes, new PR here https://github.com/openedx/frontend-app-publisher/pull/741

PR #741, opened 2022-06-27
feat: Create a Course Pane for Subscription Toggle

kiram15:
https://2u-internal.atlassian.net/browse/ENT-5834

Add behavior to toggle the Subscription Catalog inclusion boolean. This behavior should only be shown when the user has is_staff = true permissions. This also includes the work with the backend functionality of the metadata updating. This endpoint should ensure on the backend that they are staff for security purposes.

**Screenshot with info hover**
Screen Shot 2022-06-21 at 5 03 15 PM

**Error state**
Screen Shot 2022-06-21 at 5 03 48 PM

### DIFF
--- a/src/components/SidePanes/CatalogInclusionPane.jsx
+++ b/src/components/SidePanes/CatalogInclusionPane.jsx
@@ -1,0 +1,91 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { Form, Spinner } from '@edx/paragon';
+import { PUBLISHED, UNPUBLISHED } from '../../data/constants';
+
+import DiscoveryDataApiService from '../../data/services/DiscoveryDataApiService';
+import Pane from './Pane';
+
+const CatalogInclusionPane = ({
+  courseUuid,
+  draftStatuses,
+  orgInclusion,
+  subInclusion,
+}) => {
+  const [inclusion, setInclusion] = useState(subInclusion);
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState(null);
+
+  const message = 'Self-paced courses from participating partner organizations can be included in the subscription catalog.';
+  const stateError = 'Edits are not allowed while all course runs are in review.';
+  const organizationInclusionMessage = 'Organization is not currently a participating partner in the subscription catalog.';
+
+  const toggleInclusion = async (e) => {
+    e.preventDefault();
+    setIsLoading(true);
+    let draft = null;
+    if (draftStatuses.includes(PUBLISHED)) {
+      draft = false;
+    } else if (draftStatuses.includes(UNPUBLISHED)) {
+      draft = true;
+    } else {
+      // if it hits this condition, the only course runs are in review, so we will block
+      // edits at the course level
+      setIsLoading(false);
+      setError(stateError);
+    }
+
+    if (draft !== null) {
+      const data = {
+        uuid: courseUuid,
+        draft,
+        enterprise_subscription_inclusion: !(inclusion),
+      };
+      try {
+        const response = await DiscoveryDataApiService.editCourse(data);
+        setInclusion(response.data.enterprise_subscription_inclusion);
+        setError(null);
+        setIsLoading(false);
+      } catch (err) {
+        const errorText = `Unable to toggle attribute, recieved error: ${err.response.status} ${err.response.statusText}`;
+        setIsLoading(false);
+        setError(errorText);
+      }
+    }
+  };
+
+  return (
+    <Pane className="mt-1" title="Enterprise Subscriptions" info={message}>
+      <div className="font-weight-bold">Course Inclusion Status</div>
+      <Form className="mt-4">
+        <Form.Switch
+          disabled={!orgInclusion}
+          checked={inclusion}
+          className="course-inclusion-switch"
+          onChange={toggleInclusion}
+          helperText={error !== null && (<span>{error}</span>)}
+        >Included
+        </Form.Switch>
+        {!orgInclusion && (<div className="text-gray-300 x-small">{organizationInclusionMessage}</div>)}
+        {isLoading && (
+          <Spinner
+            data-testid="course-inclusion-loading"
+            animation="border"
+            variant="primary"
+            className="mr-3 float-right"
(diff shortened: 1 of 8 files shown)


azanbinzahid:
- Can you reference the backend API/Ticket?
- While testing this locally, I noticed a behavior -- Toggle this true, refresh webpage, it becomes false again.

src/components/SidePanes/CatalogInclusionPane.jsx
```diff
@@ -0,0 +1,82 @@
...
+    if (draft !== null) {
+      const data = {
+        uuid: courseUuid,
+        draft,
+        enterprise_subscription_inclusion: !(inclusion),
+      };
+      try {
+        await DiscoveryDataApiService.editCourse(data);
+        setInclusion(!(inclusion));
+        setIsLoading(false);
```
DawoudSheraz, line 40: if the API call is successful, setError(false) should be called. Otherwise, if prev API call had an error, the message will remain there despite the next call being successful.

src/components/SidePanes/CatalogInclusionPane.jsx
```diff
@@ -0,0 +1,82 @@
...
+  const toggleInclusion = async (e) => {
+    e.preventDefault();
+    setIsLoading(true);
+    let draft = null;
+    if (draftStatuses.includes(PUBLISHED)) {
+      draft = false;
+    }
+    if (draftStatuses.includes(UNPUBLISHED)) {
+      draft = true;
+    }
```
DawoudSheraz, line 25: Putting the publisher's draft flag behavior reference for Course and Course runs below, cause I suspect we would need to update this check.
- https://github.com/openedx/frontend-app-publisher/blob/master/src/components/EditCoursePage/index.jsx#L196
- https://github.com/openedx/frontend-app-publisher/blob/master/src/components/EditCoursePage/index.jsx#L117

src/components/SidePanes/CatalogInclusionPane.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+    if (draft !== null) {
+      const data = {
+        uuid: courseUuid,
+        draft,
+        enterprise_subscription_inclusion: !(inclusion),
+      };
+      try {
+        await DiscoveryDataApiService.editCourse(data);
+        setInclusion(!(inclusion));
```
DawoudSheraz, line 39: This can be misleading. We need to get the value from the response and set that value instead of just toggling.

src/components/SidePanes/CatalogInclusionPane.test.jsx
```diff
@@ -0,0 +1,23 @@
...
+      subInclusion={mockInclusion}
+    />);
+    wrapper.find(CatalogInclusionPane);
+    const title = wrapper.find('Enterprise Subscriptions');
+    expect(title);
+
+    const toggle = wrapper.find('.pgn__form-switch-input');
+    toggle.simulate('change', { target: { checked: false } });
+    expect(toggle.props().checked).toBe(false);
+  });
```
DawoudSheraz, line 22: This should have more test cases, such as draft behavior, error messages, toggle status, etc.

src/components/SidePanes/CatalogInclusionPane.jsx
```diff
@@ -0,0 +1,82 @@
...
+        enterprise_subscription_inclusion: !(inclusion),
+      };
+      try {
+        const response = await DiscoveryDataApiService.editCourse(data);
+        setInclusion(response.data.enterprise_subscription_inclusion);
+        setError(null);
+        setIsLoading(false);
+      } catch (err) {
+        setIsLoading(false);
+        setError(errorMessage);
```
DawoudSheraz, line 46: Re-iterating, showing the error from caught exception would be more helpful than showing a default error.

src/components/SidePanes/CatalogInclusionPane.jsx
```diff
@@ -0,0 +1,82 @@
...
+    let draft = null;
+    if (draftStatuses.includes(PUBLISHED)) {
+      draft = false;
+    } else if (draftStatuses.includes(UNPUBLISHED)) {
+      draft = true;
+    } else {
+      // if it hits this condition, the only course runs are in review, so we will block
+      // edits at the course level
+      setIsLoading(false);
+      setError(stateError);
```
DawoudSheraz, line 30: Shouldn't this check if the org has an inclusion flag set to True? Or hide/disable the pane if the org has opted out of sub catalog?

DawoudSheraz (commented):
- Left a few small comments
- Please reference the backend PR in PR description for other reviewers to review this with ease.

DawoudSheraz:
Including @AliAdnanSohail for additional eyes here.

src/components/SidePanes/CatalogInclusionPane.jsx
```diff
@@ -0,0 +1,91 @@
...
+  draftStatuses,
+  orgInclusion,
+}) => {
+  const [inclusion, setInclusion] = useState(subInclusion);
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState(null);
+
+  const message = 'Self-paced courses from participating partner organizations can be included in the subscription catalog.';
+  const stateError = 'Edits are not allowed while all course runs are in review.';
+  const incMessage = 'Organization is not currently a participating partner in the subscription catalog.';
```
DawoudSheraz, line 21: nit: organizationInclusionMessage

DawoudSheraz (approved):
I have not run this on local this time around but the changes look good. Please rebase and squash the commits.

src/components/SidePanes/index.jsx
```diff
@@ -59,6 +79,8 @@ SidePanes.propTypes = {
   organizationRoles: PropTypes.shape(),
   organizationUsers: PropTypes.shape(),
   removeCourseEditor: PropTypes.func,
+  enterpriseSubscriptionInclusion: PropTypes.bool.isRequired,
```
AliAdnanSohail, line 82: nit: Please order the new props alphabetically

AliAdnanSohail (approved):
Just few nits. Other than that seems good to me

PR #759, opened 2022-07-20
WS-3202: add value per lead fields

bseverino:
Adds the value per lead fields introduced in https://github.com/openedx/course-discovery/pull/3530 (`in_year_value.per_lead_usa` and `in_year_value.per_lead_international`). These fields should only be visible and able to be edited by administrators.

Internal ticket: https://2u-internal.atlassian.net/browse/WS-3202

Screen Shot 2022-07-20 at 4 47 59 PM


noisecapella (approved):

src/components/EditCoursePage/index.jsx
```diff
@@ -241,9 +254,19 @@ class EditCoursePage extends React.Component {
...
       enterprise_subscription_inclusion: courseData.enterprise_subscription_inclusion,
     };
+
     if (courseData.course_type === EXECUTIVE_EDUCATION_SLUG) {
       formattedCourseData.additional_metadata = this.formatAdditionalMetadataFields(courseData);
     }
+
+    // Check for values to prevent a new entry being created unnecessarily
+    if (
+      !!courseData.in_year_value
```
Ali-D-Akbar, line 264: nit: You can use Boolean instead of `!!`

Ali-D-Akbar (approved):

azanbinzahid:
There were 4 new values introduced and this PR adds two, when do we have plan to add other two?

bseverino:
> There were 4 new values introduced and this PR adds two, when do we have plan to add other two?

@azanbinzahid Yes, @srwang is working on the ticket to add the two other fields for value per click.

ansabgillani (approved):

PR #761, opened 2022-08-04
refactor: replaced deprecated Paragon form components

abdullahwaheed:
### Ticket
[Migrate off deprecated Paragon components](https://github.com/openedx/frontend-wg/issues/102#)

### What has changed
Updated deprecated `InputText` to `Form.Control` and `InputSelect` to `Form.Control`.

###
**Important**: We have to use ReduxForm validations in place of HTML validations to get proper validation errors provided by Paragon

PR #764, opened 2022-08-11
WS-3120: add location restriction fields

srwang:
This PR adds the location restriction fields `restriction_type`, `countries` and `states`. All field options come from course discovery.

These fields specify the regions for which a course will be made available.

~~
Note: I pulled in the most up-to-date react select component to use for these new fields. This is because the existing InputSelect doesn't have a multiselect option.  I had made a PR to add multiselect functionality to InputSelect but realized that we will be moving off InputSelect in Publisher soon.

So this PR puts the code in an interim state where there are *two* versions of the select component - the soon-to-be-deprecated InputSelect that we are using across Publisher and a new paragon component that I've added for the location restriction fields only. However [this PR](https://github.com/openedx/frontend-app-publisher/pull/761/files) will move all of Publisher to the new components and I think we should reduce the two select components to one as a part of that work.

~~

Screen Shot 2022-08-11 at 2 36 42 PM

### DIFF
--- a/src/components/EditCoursePage/EditCourseForm.jsx
+++ b/src/components/EditCoursePage/EditCourseForm.jsx
@@ -16,6 +16,9 @@ import FieldLabel from '../FieldLabel';
 import ImageUpload from '../ImageUpload';
 import RenderInputTextField from '../RenderInputTextField';
 import RenderSelectField from '../RenderSelectField';
+// TODO: remove RenderSelectFieldNew when migrating off deprecated Paragon components,
+// i.e. as a part of https://github.com/openedx/frontend-app-publisher/pull/761
+import RenderSelectFieldNew from '../RenderSelectField/updated-paragon-component';
 import RichEditor from '../RichEditor';
 import Pill from '../Pill';
 import Collapsible from '../Collapsible';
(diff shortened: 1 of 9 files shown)


ansabgillani (approved):

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -236,6 +239,12 @@ export class BaseEditCourseForm extends React.Component {
       && parseOptions(courseRunOptionsData.content_language.choices));
     const programOptions = (courseRunOptionsData
       && parseOptions(courseRunOptionsData.expected_program_type.choices));
+    const locationCountryOptions = courseOptionsData
+      && parseOptions(courseOptionsData.location_restriction.children.countries.child.choices);
```
AliAdnanSohail, line 243: Just to confirm, As this is too much nesting so do we need to add `&` to avoid the code break?

AliAdnanSohail (approved):

src/components/EditCoursePage/EditCourseForm.jsx
srwang: Do you mean something like
`parseOptions(courseOptionsData?.location_restriction.children.countries.child.choices)`?

if the argument returns false parseOptions actually breaks

AliAdnanSohail: Oh sorry I meant `?` not `&`. And if parseOptions breaks on undefined then this is good

PR #769, opened 2022-08-12
Cpappas/variant id in publisher less strict

christopappas:


src/components/EditCoursePage/__snapshots__/AdditionalMetadataFields.test.jsx.snap
```diff
@@ -79,6 +79,22 @@ exports[`AdditionalMetadata Fields Display all fields 1`] = `
...
+          className=""
+          extraText=""
+          helpText=""
+          id="variant_id.label"
+          optional={false}
+          text="Course Variant Id"
+        />
+      }
+      name="additional_metadata.variant_id"
+      required={false}
```
christopappas, line 96: note here the filed is no longer required

src/components/EditCoursePage/index.jsx
```diff
@@ -171,6 +171,7 @@ class EditCoursePage extends React.Component {
   }
 
   formatAdditionalMetadataFields(courseData) {
+    const variant_id = courseData.additional_metadata.variant_id ? courseData.additional_metadata.variant_id : null;
```
christopappas, line 174: note here that if the field is left blank, we will pass along `null` in the payload to discovery instead of an empty string (which is what we want)

src/components/EditCoursePage/AdditionalMetadataFields.jsx
```diff
@@ -44,6 +44,13 @@ function AdditionalMetadataFields(props) {
           disabled={disabled}
           required
         />
+        <Field
+          name="additional_metadata.variant_id"
+          component={RenderInputTextField}
+          label=<FieldLabel id="variant_id.label" text="Course Variant Id" />
+          disabled={disabled}
+          required={false}
```
adamstankiewicz, line 52: nit: If `required` isn't passed, it's implicitly not required, so this line is technically unnecessary, though can't hurt to be explicit I guess 🤷‍♂️

iloveagent57 (approved):
LGTM

src/components/EditCoursePage/index.jsx
adamstankiewicz, line 174: nit: camelCase the variable name and the conditional could probably be simplified a bit.

```jsx
const variantId = courseData.additional_metadata.variant_id || null;
return {
  // ...
  variant_id: variantId,
}
```

adamstankiewicz (approved):

DawoudSheraz (approved):

src/components/EditCoursePage/AdditionalMetadataFields.jsx
ansabgillani: +1 to this.

src/components/EditCoursePage/__snapshots__/AdditionalMetadataFields.test.jsx.snap
ansabgillani: Same as before. Required is by default a false field.

ansabgillani (approved):

PR #771, opened 2022-08-23
Revert "WS-3120: add location restriction fields"

srwang:
Reverts openedx/frontend-app-publisher#764

The issue was that location restriction didn't accept a null selection.


DawoudSheraz (approved):
Reviewing to unblock the pipeline.

ansabgillani (approved):
Approval to unblock the publisher.

PR #788, opened 2022-10-27
chore: fix dependency issues

ansabgillani:


openedx-webhooks:
Although this pull request is already closed, I'm still watching it for updates.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #792, opened 2022-11-16
feat: Add new course term in publisher.

zxcs001:
https://2u-internal.atlassian.net/browse/WS-3010

Frontend change for the publisher, add a new field for additional information, and course term override.

### DIFF
--- a/src/components/EditCoursePage/AdditionalMetadataFields.jsx
+++ b/src/components/EditCoursePage/AdditionalMetadataFields.jsx
@@ -51,6 +51,13 @@ function AdditionalMetadataFields(props) {
           disabled={disabled}
           required={false}
         />
+        <Field
+          name="additional_metadata.course_term_override"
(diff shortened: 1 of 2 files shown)


AliAdnanSohail (approved):

ansabgillani (approved):

src/components/EditCoursePage/AdditionalMetadataFields.jsx
```diff
@@ -51,6 +51,13 @@ function AdditionalMetadataFields(props) {
           disabled={disabled}
           required={false}
         />
+        <Field
+          name="additional_metadata.course_term_override"
+          component={RenderInputTextField}
+          label=<FieldLabel id="course_term_override.label" text="Override the course term" />
```
DawoudSheraz, line 57: nit: text should be "Course Term Override". Also add its info on field info tip to explain what that actually is.

PR #797, opened 2022-11-22
feat: hide 2U specific fields

uzairr:
Few 2U specific fields are appearing for all types of the course(s) which is creating confusion for the partners.Due to this change, 2U fields would only be displayed for 2U courses

PROD-3032

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,5 @@
 NODE_ENV='production'
+IDENTIFIER_2U='2U'
 BASE_URL=null
 LMS_BASE_URL=null
 STUDIO_BASE_URL=null
--- a/.env.development
+++ b/.env.development
@@ -1,4 +1,5 @@
 NODE_ENV='development'
+IDENTIFIER_2U='2U'
 PORT=18400
 BASE_URL='localhost:18400'
 LMS_BASE_URL='http://localhost:18000'
--- a/.env.test
+++ b/.env.test
@@ -5,3 +5,4 @@ LOGO_TRADEMARK_URL=https://edx-cdn.org/v3/default/logo-trademark.svg
 LOGO_WHITE_URL=https://edx-cdn.org/v3/default/logo-white.svg
 FAVICON_URL=https://edx-cdn.org/v3/default/favicon.ico
 SITE_NAME='edX'
+IDENTIFIER_2U='2U'
--- a/src/components/EditCoursePage/EditCourseForm.jsx
+++ b/src/components/EditCoursePage/EditCourseForm.jsx
@@ -6,6 +6,7 @@ import {
 import { Link } from 'react-router-dom';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
+import { getConfig } from '@edx/frontend-platform';
 import { getAuthenticatedUser } from '@edx/frontend-platform/auth';
 import { Icon, Hyperlink, Chip } from '@edx/paragon';
 import { CloseSmall } from '@edx/paragon/icons';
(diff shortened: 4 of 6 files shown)


arbrandes (changes requested):
Can we please make this a configurable setting, instead of making hard-coded references to 2U in the code?

Ali-D-Akbar (approved):
Do we have a snapshot where these override fields appear? 
If yes then it looks good to me.

AliAdnanSohail (approved):

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -226,6 +227,10 @@ export class BaseEditCourseForm extends React.Component {
     this.setCollapsible(true);
   }
 
+  filterCourseTypes2U(courseType) {
```
arbrandes, line 230: If we remove "2U" from the object and parameter names we can avoid confusion in the future, making this a truly generic feature.

```suggestion
  filterCourseTypes(courseType) {
```

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -306,6 +311,7 @@ export class BaseEditCourseForm extends React.Component {
       priceLabels,
       runTypeModes,
     } = parsedTypeOptions;
+    const courseTypes2U = courseTypeOptions.filter(this.filterCourseTypes2U);
```
arbrandes, line 314:
```suggestion
    const courseTypesToFilter = courseTypeOptions.filter(this.filterCourseTypes);
```

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -1046,34 +1052,38 @@ export class BaseEditCourseForm extends React.Component {
...
-              optional
-            />
-            <Field
-              name="organization_logo_override_url"
-              component={ImageUpload}
-              label={(
-                <FieldLabel
-                  id="organization_logo_override.label"
-                  text="Organization Logo Override"
+            {courseTypes2U?.length > 0 && (
```
arbrandes, line 1055:
```suggestion
            {courseTypesToFilter?.length > 0 && (
```

.env
```diff
@@ -1,4 +1,5 @@
 NODE_ENV='production'
+IDENTIFIER_2U='2U'
```
arbrandes, line 2: Let's rename this as well, and remove the "2U" default setting which does not make sense for all use cases.

```suggestion
COURSE_TYPE_FILTER=''
```

.env.development
```diff
@@ -1,4 +1,5 @@
 NODE_ENV='development'
+IDENTIFIER_2U='2U'
```
arbrandes, line 2:
```suggestion
COURSE_TYPE_FILTER=''
```

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -226,6 +227,10 @@ export class BaseEditCourseForm extends React.Component {
     this.setCollapsible(true);
   }
 
+  filterCourseTypes2U(courseType) {
+    return courseType.label.includes(getConfig().IDENTIFIER_2U);
```
arbrandes, line 231:
```suggestion
    return courseType.label.includes(getConfig().COURSE_TYPE_FILTER);
```

.env.test
```diff
@@ -5,3 +5,4 @@ LOGO_TRADEMARK_URL=https://edx-cdn.org/v3/default/logo-trademark.svg
 LOGO_WHITE_URL=https://edx-cdn.org/v3/default/logo-white.svg
 FAVICON_URL=https://edx-cdn.org/v3/default/favicon.ico
 SITE_NAME='edX'
+IDENTIFIER_2U='2U'
```
arbrandes, line 8:
```suggestion
COURSE_TYPE_FILTER=''
```

arbrandes (changes requested):
Still think we can make this more generic.  Thanks for bearing with me.

ansabgillani (approved):

PR #815, opened 2023-01-20
feat: refactor Staffer Form organization field to a searchable dropdown

zawan-ila:
# [PROD-3097](https://2u-internal.atlassian.net/browse/PROD-3097)

## Description

This PR updates the Staffer Form organization field from a text field to a searchable dropdown

### Before
Screenshot 2023-01-21 at 1 36 25 AM

### After
Screenshot 2023-01-21 at 1 35 55 AM


src/components/StafferPage/OrganizationDropdown.jsx
```diff
@@ -0,0 +1,65 @@
+import PropTypes from 'prop-types';
+import CreatableSelect from 'react-select/creatable';
+import { Alert } from '@edx/paragon';
+
+const OrganizationDropdown = props => {
+  const {
+    label, input, options, meta: { touched, error },
+  } = props;
+
```
DawoudSheraz, line 9: Did we not create a similar component for UserPane's course editors? I suggest refactoring the common component and using that common component in both org dropdown and user pane.

DawoudSheraz: @zawan-ila

zawan-ila: We did not create a component for UserPane. We directly used a component provided by react-select [(code ref)](https://github.com/openedx/frontend-app-publisher/blob/master/src/components/SidePanes/UsersPane.jsx#L166)

We did create a component for course tags, however, that component is somewhat different from our current component, in particular, that it fetches options async and renders a different component from react-select than our current one. On top of that, the rendered components(from react-select) take in some significantly different props. As such, while we can still indeed combine both these inside a single component, i am unsure of the benefits of such an approach. 

Please have a look at both of these components ( [Course Tags](https://github.com/openedx/frontend-app-publisher/blob/master/src/components/ReduxFormCreatableSelect/index.jsx#L7) vs [Staffer Organization](https://github.com/openedx/frontend-app-publisher/blob/caeda90867b6ad65e68d90b832e6fe548490ce5f/src/components/StafferPage/OrganizationDropdown.jsx#L5) ) and let me know what you think.

DawoudSheraz: > As such, while we can still indeed combine both these inside a single component, i am unsure of the benefits of such an approach.

The benefit is re-usability. Both components are doing similar things. I would not term that as a blocker but having a dedicated component, ideally in openedx paragon, would be beneficial for the community.

zawan-ila: have refactored reduxformcreatableselect to accommodate our usecases

ansabgillani (approved):

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -508,21 +494,23 @@ export class BaseEditCourseForm extends React.Component {
...
-                Array.isArray(courseTags)
-                  ? this.courseTagsToSelectValues(courseTags)
-                  : null
-              }
-              setValue={(newValue) => setCourseTags(newValue)}
+              isCreatable
+              styles={{
+                menu: styles => ({ ...styles, zIndex: 5 }),
+                control: styles => ({ ...styles, marginTop: '8px' }),
+              }}
```
Ali-D-Akbar, line 505: Can we move the styles in EditCourseForm.scss?

Ali-D-Akbar (commented):
Is there any visual change regarding tag field in edit course? If there is one, can you add a screenshot of it as well?

Also, left a comment before approval.

src/components/EditCoursePage/EditCourseForm.jsx
DawoudSheraz: +1 on this.

DawoudSheraz (approved):
I will test the changes on stage.

src/components/EditCoursePage/EditCourseForm.jsx
zawan-ila: also this PR does not change tag field in any visual way

Ali-D-Akbar (approved):

PR #859, opened 2023-03-22
feat: remove some field required validation on the basis of external course marketing type

AliAdnanSohail:
[Remove/Override Certificate data validation for EE Sprints and Stacks](https://2u-internal.atlassian.net/browse/PROD-3259)


Ali-D-Akbar (approved):

src/components/EditCoursePage/AdditionalMetadataFields.jsx
```diff
@@ -9,13 +9,18 @@ import DateTimeField from '../DateTimeField';
...
-    disabled, sourceInfo,
+    disabled, sourceInfo, externalCourseMarketingType,
   } = props;
   const sourceSlug = sourceInfo?.slug;
 
   const SOURCE_SLUG_REQUIRED_FIELDS = JSON.parse(process.env.ADDITIONAL_METADATA_REQUIRED_FIELDS);
 
   function isRequiredField(fieldName) {
+    if (['certificate_info_heading', 'certificate_info_blurb'].includes(fieldName)) {
+      if (['sprint', 'course_stack'].includes(externalCourseMarketingType)) {
```
DawoudSheraz, line 20: I would advocate setting up `['sprint', 'course_stack']` as an environment variable and use edx-internal to setup these instead of hard-coding these values

AliAdnanSohail: @DawoudSheraz ideally these `certificate` fields should also be moved in environment variable. But this is just a single case for a field for some course types. And It was not making sense to move both arrays in separate ENV variable. If you have something in your mind how we can categorize this and named it properly please let me know. I'll update.

DawoudSheraz: Hmmm, for now, it is ok. We will need a similar solution like `ADDITIONAL_METADATA_REQUIRED_FIELDS` if we need more breakdown of required fields based on `externalCourseMarketingType`.

PR #864, opened 2023-04-10
refactor: refactor formatting of data before sending to the backend

zawan-ila:
[PROD-3202](https://2u-internal.atlassian.net/browse/PROD-3202)

## Description

In this PR, we make some changes to the formatting of data before sending it to the backend. This is done so as to ensure that the backend receives data in the format that it is expecting. This is required to enable the backend to decide whether the incoming data is different than the one present in the database or are there any real changes?

### DIFF
--- a/src/components/EditCoursePage/index.jsx
+++ b/src/components/EditCoursePage/index.jsx
@@ -250,6 +250,30 @@ class EditCoursePage extends React.Component {
     return editedRun;
   }
 
+  formatFacts(addMeta) {
+    const fact_1 = {
+      heading: addMeta.facts_1_heading,
+      blurb: addMeta.facts_1_blurb,
+    };
+    const fact_2 = {
+      heading: addMeta.facts_2_heading,
+      blurb: addMeta.facts_2_blurb,
+    };
+
+    if (fact_1.heading && fact_2.heading) {
+      return [
+        fact_1, fact_2,
+      ];
+    }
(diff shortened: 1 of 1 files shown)


src/components/EditCoursePage/index.jsx
```diff
@@ -298,32 +313,18 @@ class EditCoursePage extends React.Component {
...
-        geoLocationLat: courseData.geolocation.lat,
-      };
-    }
     return {
-      geoLocationName: null,
-      geoLocationLng: null,
-      geoLocationLat: null,
+      geoLocationName: courseData.geolocation?.location_name || '',
+      geoLocationLng: courseData.geolocation?.lng || null,
+      geoLocationLat: courseData.geolocation?.lat || null,
```
DawoudSheraz, line 327: This is backend data rendering on publisher. Are you sure it is working as expected, like adding [] for fields?

src/components/EditCoursePage/index.jsx
```diff
@@ -250,6 +250,30 @@ class EditCoursePage extends React.Component {
...
+      ];
+    }
+    if (fact_1.heading) {
+      return [fact_1];
+    }
+    if (fact_2.heading) {
+      return [fact_2];
+    }
+    return [];
+  }
```
DawoudSheraz, line 275: Either both facts should be there or none. Only one fact is not allowed.

PR #869, opened 2023-05-12
fix: add redux change to update fields changed by backend

DawoudSheraz:
### Description

In certain cases of edits, the Save & Re-publish would not change to the disabled "Changes Saved" button after the API call has been completed. This happens when the backend has changed the form values in one way or another in the returned response. On publisher, if the form values do not match returned values, it treats as if the edits are still in progress. That is also why refreshing the page fixed the button because form values are populated with the latest DB data returned by API. 
This PR adds a few fields, that can be edited by backend, to sync their data with form data via a redux change to ensure the button moves to the correct state and provide feedback to the editor that the operation was successful.

### Testing
- Run publisher on your local. Make sure studio, ecommerce, and discovery are also running.
- Open a course that has values against most fields.
- Set url_slug to empty, remove focus, and press "Save & Re-publish". Notice the slug field gets populated with data
- Add/Remove topics and repeat. Ensure the order of tags is the same.
- Add values for 1-2 fields for In Year value section. Ensure the default values from the backend are populated for other fields after saving.
- Under transcript lang, add English and then add Arabic. After save, Arabic will be above English and button will change.

### DIFF
--- a/src/data/actions/courseInfo.test.js
+++ b/src/data/actions/courseInfo.test.js
@@ -20,6 +20,7 @@ import {
   editCourseFail,
   clearCourseSaved,
   setCreateAlertOff,
+  updateFormValuesAfterSave,
 } from './courseInfo';
 import * as types from '../constants/courseInfo';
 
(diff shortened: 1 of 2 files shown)


src/data/actions/courseInfo.test.js
```diff
@@ -187,6 +188,43 @@ describe('courseInfo edit course actions', () => {
...
+    expect(changeMock).toHaveBeenNthCalledWith(3, 'geoLocationLng', '45.0000');
+    expect(changeMock).toHaveBeenNthCalledWith(4, 'tags', ['tag1', 'tag2']);
+    expect(changeMock).toHaveBeenNthCalledWith(5, 'url_slug', 'test_slug');
+    expect(changeMock).toHaveBeenNthCalledWith(6, 'in_year_value.per_lead_usa', 10);
+    expect(changeMock).toHaveBeenNthCalledWith(7, 'in_year_value.per_lead_international', 20);
+    expect(changeMock).toHaveBeenNthCalledWith(8, 'in_year_value.per_click_usa', 25);
+    expect(changeMock).toHaveBeenNthCalledWith(9, 'in_year_value.per_click_international', 30);
+    expect(changeMock).toHaveBeenNthCalledWith(10, 'course_runs[0].status', 'published');
+    expect(changeMock).toHaveBeenNthCalledWith(11, 'course_runs[0].transcript_languages', ['en-us']);
+  });
```
Ali-D-Akbar, line 227: nit: not a blocker but this can be reduced to a for-loop iterating over `formValues`.

Ali-D-Akbar (approved):
Working fine for me.

uzairr (approved):

src/data/actions/courseInfo.test.js
DawoudSheraz: It required a JSON to map fields -> values and I was being a bit lazy with it, so I went with this approach.

PR #873, opened 2023-05-15
fix: tiny mce button find error

AfaqShuaib09:
## [PROD-3225](https://2u-internal.atlassian.net/browse/PROD-3255)

This PR fixes TinyMCE editor Language Selector error 
``` failed to find and set button ID in TinyMceLanguageSelectorPlugin after 3 attempts.```
The constant var `showMarketingFields` is used to show/hide the marketing fields in the course editor page but it was initialized with 
```
 const showMarketingFields = !currentFormValues.type || !courseTypes[currentFormValues.type]
      || courseTypes[currentFormValues.type].course_run_types.some((crt) => crt.is_marketable);
```
which is true if currentFormValues.type is null or undefined until CurrentFormValues is loaded from the server. So, the marketing fields were appearing until the CurrentFormValues is loaded from the server. This caused an issue where the TinyMCE language selector console-logged an error stating that it was unable to find the button. This occurred because the field was shown first and then removed based on the value of `showMarketingFields`, causing the TinyMCE language selector to encounter an error.

image

------

### Testing Instructions:
- Create a Masters Only Course through publisher
- Go to the course editor page
- Verify that the marketing fields are not appearing
- Check from developer tools console that there is no error related to TinyMCE editor

### CODE REVIEW

File: src/components/EditCoursePage/EditCourseForm.jsx [hunk 1]
```diff
@@ -283,8 +283,8 @@ export class BaseEditCourseForm extends React.Component {
       runTypeModes,
     } = parsedTypeOptions;
     const disabled = courseInReview || !editable;
-    const showMarketingFields = !currentFormValues.type || !courseTypes[currentFormValues.type]
-      || courseTypes[currentFormValues.type].course_run_types.some((crt) => crt.is_marketable);
+    const showMarketingFields = (currentFormValues.type && courseTypes[currentFormValues.type]) === undefined ? false
+      : courseTypes[currentFormValues.type].course_run_types.some((crt) => crt.is_marketable);
```
DawoudSheraz, line 287: nit: convert this to a function and add comment/jsdoc to explain the criteria for displaying marketing fields.

DawoudSheraz (approved):

PR #883, opened 2023-06-14
feat: update hardcode preview_url for courses having updated slug

AfaqShuaib09:
This PR updates `Preview URL` for courses having new slug format


src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -139,7 +139,8 @@ export class BaseEditCourseForm extends React.Component {
       return (
         <>
           <Hyperlink
-            destination={`https://www.edx.org/preview/course/${courseInfo.data.url_slug}`}
+            name="preview-url"
+            destination={`https://www.edx.org/preview/${courseInfo?.data?.url_slug.includes('/') ? '' : 'course/'}${courseInfo.data.url_slug}`}
```
DawoudSheraz, line 143: Please remove hardcoded edx.org/preview and move this to a setting/env var. Never realized this was hardcoded.

AfaqShuaib09: fixed.
`edx-internal` PR: https://github.com/edx/edx-internal/pull/8625

src/components/EditCoursePage/EditCourseForm.test.jsx
```diff
@@ -244,6 +245,72 @@ describe('BaseEditCourseForm', () => {
...
+      entitlement={{ sku: 'ABC1234' }}
+      courseStatuses={[REVIEWED]}
+      courseInfo={courseInfoWithUrlSlug}
+      courseOptions={courseOptions}
+      courseRunOptions={courseRunOptions}
+      uuid={initialValuesFull.uuid}
+      type={initialValuesFull.type}
+      id="edit-course-form"
+    />);
+    expect(wrapper.instance().getLinkComponent([REVIEWED], courseInfoWithUrlSlug)).toEqual(
```
zawan-ila, line 300: You wish to assert something about `wrapper`. You do that by calling a method(`getLinkComponent`) on `wrapper` that does not depend on `wrapper` but merely on its arguments. So essentially `wrapper` is irrelevant in this test and you are just testing `getLinkComponent`. If this is indeed the case we should remove `wrapper` and call `getLinkComponent` directly?

AfaqShuaib09: It's a class method and can't be called independently, IMO.
> In React, class methods are typically bound to component instances and are meant to be invoked as part of the component lifecycle or in response to specific events. They are not designed to be tested independently like regular functions.

zawan-ila: Hmm.
I feel it is better to restructure these tests to use something like `wrapper.find` to get the `Hyperlink` and then assert on the `props`, rather than calling `getLinkComponent`.

zawan-ila (approved):

src/components/EditCoursePage/EditCourseForm.test.jsx
DawoudSheraz: I would suggest writing these tests with rtl instead of enzmye. Enzyme is going to be out of fashion, so let's move to RTL with a new code instead of doing that work later.

DawoudSheraz: And I also 2nd @zawan-ila  that rather than asserting on entire component, verify its rendering and props value that are important in this unit test.

AfaqShuaib09: > I would suggest writing these tests with rtl instead of enzmye. Enzyme is going to be out of fashion, so let's move to RTL with a new code instead of doing that work later.

Testing Redux Form in RTL can be more complex compared to enzyme. It requires managing the store and providers to maintain the form's state during testing. Additionally, moving a few tests from enzyme to RTL may introduce confusion. It might be better to address this in a follow-up pull request where we can focus on properly integrating RTL with Redux Form.

DawoudSheraz: What's the complexity value? Any links/refs on that? If it it indeed high effort, we can do it later. But if it is quick,  lets start from this PR.

AfaqShuaib09: We need to setup Stores and Providers for the tests just like in the following links. 

Ref Links:
https://stackoverflow.com/questions/46932791/redux-form-test-using-jest
https://dev.to/ip4422/redux-form-and-typescript-testing-with-react-testing-library-5bla

PR #885, opened 2023-06-27
fix: regex pattern for slug pattern

AfaqShuaib09:
Description
---------------
This PR fixes regex pattern for URL slug field.

### Testing Instructions:
- Enter the current url slug pattern (Alphanumeric, -, _) and check if it is working properly, Enter the invalid slugs as well to check the validity
- Update `IS_NEW_SLUG_FORMAT_ENABLED` value from `.env.development` file to true and check if it is accepting both updated & old url slug for OCM courses if none of its course in `POST REVIEW` or `IN REVIEW` state and old format slug for remaining


DawoudSheraz (approved):
publisher changes are working for me, but due to backend validation checks in course api, the slugs were not updated.

PR #886, opened 2023-07-03
feat: allow both slug formats until complete rollout

zawan-ila:
[PROD-3449](https://2u-internal.atlassian.net/browse/PROD-3449)
------------------
This PR temporarily allows normal slug format even if `learn` slug format is enabled

### Testing Instructions:
- Enter the current url slug pattern (Alphanumeric, -, _) and check if it is working properly. Enter invalid slugs and check appropriate message is displayed.
- Update `IS_NEW_SLUG_FORMAT_ENABLED` value from `.env.development` file to true and check if it is accepting both updated & old url slug for OCM courses and only old url format for other courses.  **Check that behavior for OCM courses is same in case of both published and unpublished courses**

### Note: 
You may get validation errors on the backend with `learn` format slugs. You can either ignore them or comment out [this line in discovery](https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/api/v1/views/courses.py#L390)


AfaqShuaib09 (approved):

DawoudSheraz (approved):
I have not tested it out on local. The changes look good.

PR #888, opened 2023-07-12
fix: refresh course data when moving to review to get automated generated slug

DawoudSheraz:
### Description
The learn/subdirectory slug for a course is generated when a course run is moved to legal review. When the action to move a course run to legal review is initiated, two API calls are made. One for Course update and then course run update. It is in 2nd API call where the course slug is updated. This updates the data on the backend but the frontend does not contain updated course information. Refreshing of the page is required to bring in fresh data. This PR adds a change to refreshData when a course run is moved to legal review, only if the new slug format flag is enabled. 

**There are no unit tests in place for functions like handleCourseRuns or editCourse. Setting their unit tests would require some effort, which is already being discussed internally. That's why I did not add a unit test.**


src/data/actions/courseInfo.js
```diff
@@ -237,6 +237,11 @@ function handleCourseRuns(
     });
     dispatch(editCourseSuccess(newCourse));
     if (submitReview) {
+      if (process.env.IS_NEW_SLUG_FORMAT_ENABLED === 'true' && !internalReview) {
+        // Refresh data to get updated url slug that has been generated by backend when a course run is moved
+        // into legal review.
+        refreshData();
```
AfaqShuaib09, line 243: Can we add testcase for this scenario in `courseInfo.test.js`?

AfaqShuaib09 (approved):
Tested on local, LGTM.

AliAdnanSohail (approved):

PR #890, opened 2023-07-19
feat: add watchers field

AfaqShuaib09:


src/utils/validation.js
```diff
@@ -15,6 +15,13 @@ function courseTagValidate(tagValue, selectValue, options) {
   return ![...selectValue, ...options].some(x => x.label.toLowerCase() === tagValue.toLowerCase());
 }
 
+function emailValidate(emailValue, selectValue, options) {
+  // emailValue should only contain alphabets, numbers, hyphen, underscore, dot and @
+  if (!/^[^\s@]+@[^\s@]+\.[^\s@]+$/i.test(emailValue)) { return false; }
+  // disallow emails that have already been selected or are present in options(dropdown)
+  return ![...selectValue, ...options].some(x => x.label.toLowerCase() === emailValue.toLowerCase());
```
DawoudSheraz, line 22: This particular validation should be on component level. This util method should only focus on email text validation.

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -1091,6 +1093,31 @@ export class BaseEditCourseForm extends React.Component {
...
+                      <p>
+                        A list of email addresses that will receive
+                        notifications when the course run of the course is published or reviewed.
+                      </p>
+                    )}
+                    optional
+                  />
+                )}
+                isMulti
+                disabled={!(courseInfo?.data?.course_run_statuses?.includes(REVIEW_BY_INTERNAL) && administrator)}
```
DawoudSheraz, line 1114: - If I remember the spec, the watchers can be added at any point (but not once scheduled/published). Double-check this.
- We should make this field prominent, either moving to the side pane or moving higher in the course form. If the side pane does not take much effort, that would be preferrable. This particular field does not impact the course content behavior.

DawoudSheraz: Also, the form remains in an uncleaned state due to the ordering of watcher fields coming from backend. The form field needs to be refreshed. It is already happening for many of the fields in Publisher.

DawoudSheraz (commented):
I have added some followup comments.

src/utils/validation.js
AfaqShuaib09: I've followed here the same convention which is currently followed for the tags validation

- https://github.com/openedx/frontend-app-publisher/blob/master/src/utils/validation.js#L11-L16
- https://github.com/openedx/frontend-app-publisher/blob/master/src/components/EditCoursePage/EditCourseForm.jsx#L511-L512

src/components/EditCoursePage/EditCourseForm.jsx
AfaqShuaib09: 1. I've doubled check this from specs doc
image
2. Ok, I'm moving this field upwards after slug field to make it prominent, adding a side pane and manage field state require some effort, can be done in some followup PR
3. Working on it

AfaqShuaib09: > Also, the form remains in an uncleaned state due to the ordering of watcher fields coming from backend. The form field needs to be refreshed. It is already happening for many of the fields in Publisher.
Done

zawan-ila (approved):

src/components/EditCoursePage/index.jsx
```diff
@@ -370,6 +370,7 @@ class EditCoursePage extends React.Component {
       organization_logo_override: courseData.organization_logo_override_url,
       organization_short_code_override:
         courseData.organization_short_code_override,
+      watchers: courseData.watchers_list ? courseData.watchers_list.map((watcher) => watcher.value) : [],
```
DawoudSheraz, line 373: wouldn't this turn watchers into empty list if a non-admin user (like Editor) publishes or edits the course?

src/utils/validation.js
DawoudSheraz: Right. Tags have a specific use-case. Email is generic, in my opinion.

src/components/EditCoursePage/index.jsx
DawoudSheraz: Try opening the course with an Editor user. I suspect the above will be true.

AfaqShuaib09: nope, it doesn't turn into an empty list I've already tried this for course Editor Role on my local. And we are not showing this field to Course Editor. Only Admin user/ PCs can view or edit this field.

#### Course Editor View:
image


#### Admin User/ PC view:
image

I've tried updating the course with course Editor role user it didn't empty the watchers field

DawoudSheraz: But we are adding key in the request data and are adding [] if the form data is not found. What's the request value in network tab when editing as an Editor user? Can you double check if we can remove watchers from a course (assuming that if [] is not changing watchers on backend, then we cannot remove all watchers for a course when needed).

AfaqShuaib09: Actually form data initialValues send to `prepareSendCourseData` method if we don't show change them, as Watchers field is not showing with `Couse Editor` role on Course Edit Form so they can't edit this field.  `prepareSendCourseData` doesn't send empty watchers field to backend because we're not editing watchers initial values. It send same previous watcher value to backend.

> (assuming that if [] is not changing watchers on backend, then we cannot remove all watchers for a course when needed).
We can empty this field when needed by editing this field with PCs/Admin Role. `[]` this is only possible when watchers value is `null` not in case of Course Editor editing the form because in that case it's sending previous value to backend

src/utils/validation.js
AfaqShuaib09: I don't find any benefit in moving it to the component level. Because we can reuse this function anywhere after importing it to that file

src/components/EditCoursePage/index.jsx
AfaqShuaib09: - https://github.com/openedx/frontend-app-publisher/pull/890/files#diff-1018929192344c42ef49da293a7c2bd439019169f1fa3bc10b84b6b2805fe999R666
Here we are setting the initial value of `watchers` coming from backend in format for multi email field

- https://github.com/openedx/frontend-app-publisher/pull/890/files#diff-1018929192344c42ef49da293a7c2bd439019169f1fa3bc10b84b6b2805fe999R666
Here we are passing this formatted value to the field (if we don't show this field its state remain unaffected)

- https://github.com/openedx/frontend-app-publisher/pull/890/files#diff-1018929192344c42ef49da293a7c2bd439019169f1fa3bc10b84b6b2805fe999R373
Here we're converting watchers back to list format

(If we hide the "watchers" field for course editors, it will not affect the "watchers_list" state. During the onSave event, the code will retain the existing state of "watchers_list" rather than passing an empty state)

src/components/EditCoursePage/index.jsx
```diff
@@ -657,9 +659,11 @@ class EditCoursePage extends React.Component {
       enterprise_subscription_inclusion,
       organization_short_code_override,
       organization_logo_override_url,
+      watchers,
```
DawoudSheraz, line 662: nit: add a comment to explain why watchers field is needed here alongside watchers_list.

PR #892, opened 2023-07-25
fix: make courses with type empty editable in publisher

zawan-ila:
### Description
Courses with type `empty` are not editable in publisher. This PR attempts to fix that.


Ali-D-Akbar (approved):

DawoudSheraz (approved):
I would suggest adding a comment above the line to ensure the context is saved in the code. This is a niche case and requires some context to understand.

PR #897, opened 2023-08-01
fix: ensure prepush lint returns error on failure

brian-smith-tcril:
After making https://github.com/openedx/frontend-app-admin-portal/pull/1014, I did a full search of the org for `; npm run` and ended up here. I updated `prepush` in `package.json` to use `&&` instead of `;` to ensure we return an error code on lint failure.


brian-smith-tcril:
Just realized this is prepush, not CI, investigating...

PR #931, opened 2023-11-22
feat: add variant ID field CourseRun form

AfaqShuaib09:
[PROD-3742]()
-------
This PR adds variant_id to CourseRunForm and make it visible only for external courses

image

### DIFF
--- a/src/components/EditCoursePage/CollapsibleCourseRun.test.jsx
+++ b/src/components/EditCoursePage/CollapsibleCourseRun.test.jsx
@@ -4,7 +4,9 @@ import { shallowToJson } from 'enzyme-to-json';
 
 import CollapsibleCourseRun from './CollapsibleCourseRun';
 import { courseSubmitRun } from '../../data/actions/courseSubmitInfo';
-import { AUDIT_TRACK, MASTERS_TRACK, VERIFIED_TRACK } from '../../data/constants';
+import {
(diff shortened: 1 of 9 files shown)


src/components/EditCoursePage/CollapsibleCourseRun.test.jsx
```diff
@@ -109,6 +111,32 @@ describe('Collapsible Course Run', () => {
...
+      courseRun={unpublishedCourseRun}
+      courseId="test-course"
+      courseUuid="11111111-1111-1111-1111-111111111111"
+      type="8a8f30e1-23ce-4ed3-a361-1325c656b67b"
+      currentFormValues={currentFormValues}
+      courseRunTypeOptions={courseRunTypeOptions}
+      index={1}
+      courseInfo={courseInfo}
+    />);
+    expect(shallowToJson(component)).toMatchSnapshot();
```
Ali-D-Akbar, line 137: Do we need to use `shallowToJson` in this example? Why can't we simply do that without `shallowToJson`?

Ali-D-Akbar (approved):

src/components/EditCoursePage/CollapsibleCourseRun.test.jsx
AfaqShuaib09: We can do it without using `shallowToJson`. I included it here because it is utilized in the other tests of the same file. 
However, I've updated the test to verify this via both (snapshot, find) methods

DawoudSheraz (approved):
I have not tested it locally but looks alright.

PR #933, opened 2023-12-18
fix: include reviewed to allowed statuses

kiram15:
https://2u-internal.atlassian.net/browse/ENT-8076

Previously, we only allowed you to toggle the enterprise subscription inclusion if your course run was published, but now we are also allowing it if it is reviewed and scheduled. In response to a bug filed by a Harvard customer.

### DIFF
--- a/src/components/SidePanes/CatalogInclusionPane.jsx
+++ b/src/components/SidePanes/CatalogInclusionPane.jsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { Form, Spinner } from '@edx/paragon';
-import { PUBLISHED, UNPUBLISHED } from '../../data/constants';
+import { PUBLISHED, REVIEWED, UNPUBLISHED } from '../../data/constants';
 
 import DiscoveryDataApiService from '../../data/services/DiscoveryDataApiService';
 import Pane from './Pane';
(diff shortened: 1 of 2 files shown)


src/components/SidePanes/CatalogInclusionPane.jsx
```diff
@@ -47,7 +47,7 @@ const CatalogInclusionPane = ({
         setError(null);
         setIsLoading(false);
       } catch (err) {
-        const errorText = `Unable to toggle attribute, recieved error: ${err.response.status} ${err.response.statusText}`;
+        const errorText = `Unable to toggle attribute, received error: ${err.response.status} ${err.response.statusText}`;
```
Ali-D-Akbar, line 50: Would be nice to have a unit test for this but case too but not a blocker.

Ali-D-Akbar (approved):

src/components/SidePanes/CatalogInclusionPane.test.jsx
```diff
@@ -22,6 +22,21 @@ describe('CatalogInclusionPane', () => {
...
+      courseUuid={mockUuid}
+      subInclusion={mockInclusion}
+      draftStatuses={['reviewed']}
+      orgInclusion
+    />);
+    wrapper.find(CatalogInclusionPane);
+    const title = wrapper.find('Enterprise Subscriptions');
+    expect(title);
+    const toggle = wrapper.find('.pgn__form-switch-input');
+    toggle.simulate('change', { target: { checked: false } });
```
AfaqShuaib09, line 36: `toggle.simulate('change', { target: { checked: false } });`

I believe it should be `true` here because the default value is `false`

AfaqShuaib09 (approved):

PR #959, opened 2024-04-22
feat: add Restriction Type in EditCoursePage

zawan-ila:
[PROD-4008](https://2u-internal.atlassian.net/browse/PROD-4008)
-------
This PR adds restriction_type field to CourseRunForm and makes it visible only for external courses

Screenshot 2024-04-22 at 10 33 54 PM

## Testing

1. Check out https://github.com/openedx/course-discovery/pull/4331
2. Open an executive education course in publisher
3. Update its restriction_type field
4. Verify that the restriction_type is appropriately updated in the backend. This can be verified through the CourseRunRestriction model in Django admin


Ali-D-Akbar (approved):

src/utils/index.js
```diff
@@ -369,6 +369,12 @@ const buildInitialPrices = (entitlements, courseRuns) => {
 const hasMastersTrack = (runTypeUuid, runTypeModes) => (!!runTypeUuid
   && !!runTypeModes[runTypeUuid] && runTypeModes[runTypeUuid].includes(MASTERS_TRACK.key));
 
+const restrictionTypeOptions = [
+  { value: 'custom-b2b-enterprise', label: 'Custom Enterprise' },
+  { value: '', label: '--------' },
```
DawoudSheraz, line 374: nit: the empty value should be first, no?

src/components/EditCoursePage/EditCoursePage.test.jsx
```diff
@@ -64,6 +65,7 @@ describe('EditCoursePage', () => {
           end: defaultEnd,
           upgrade_deadline_override: '2019-05-10T00:00:00Z',
           variant_id: null,
+          restriction_type: restrictionType,
```
DawoudSheraz, line 68: do we need this value across all test data? Shouldn't coalesce handle the missing data?

zawan-ila: This blob of data is a "mock" for the network response. Since the network response only has `null`, `custom-b2c` and `custom-b2b-enterprise` as possible values, I am testing here with both a null and string response. This is the string version and the course run below is the null version.

src/utils/index.js
zawan-ila: I'll update it before merging

src/components/EditCoursePage/EditCoursePage.test.jsx
DawoudSheraz: Got it.

DawoudSheraz (approved):
One small comment, otherwise looks good. Make sure to merge this after backend PR has been merged.

src/components/EditCoursePage/CollapsibleCourseRun.test.jsx
```diff
@@ -138,6 +138,36 @@ describe('Collapsible Course Run', () => {
...
+        courseUuid="11111111-1111-1111-1111-111111111111"
+        type="8a8f30e1-23ce-4ed3-a361-1325c656b67b"
+        currentFormValues={currentFormValues}
+        courseRunTypeOptions={courseRunTypeOptions}
+        index={1}
+        courseInfo={courseInfo}
+      />,
+    );
+    const restrictionTypeField = component.find('Field[name="test-course.restriction_type"]');
+    expect(restrictionTypeField.exists()).toBe(true);
```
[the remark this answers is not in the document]
zawan-ila: I don't see how we can do that here. In particular, I am unable to find the select component here. I think it is because of the shallow render. I tried mounting it, but then it complained about not being inside a reduxForm.

I have added a test for the actual value in `EditCoursePage.test.jsx`.

PR #964, opened 2024-06-21
fetch restricted runs in publisher

zawan-ila:
[PROD-4088](https://2u-internal.atlassian.net/browse/PROD-4088)
-------
This PR adds adds the restriction params to the API calls in Publisher to ensure that restricted runs are always fetched/updated properly in Publisher.

## Testing

1. Check out this branch
2. Open/Create an executive education course in publisher
3. Add a restriction type. Then change it. Finally remove it. 
4. Verify that the behavior is as expected.

### DIFF
--- a/src/data/constants/index.js
+++ b/src/data/constants/index.js
@@ -39,7 +39,7 @@ const RESTRICTION_TYPE_OPTIONS = [
   { value: 'custom-b2c', label: 'Custom B2C' },
 ];
 
-const RESTRICTION_TYPE_VALUES = RESTRICTION_TYPE_OPTIONS.map(opt=>opt.value).filter(Boolean).join(',')
+const RESTRICTION_TYPE_VALUES = RESTRICTION_TYPE_OPTIONS.map(opt => opt.value).filter(Boolean).join(',');
 
 export {
   VERIFIED_TRACK,
--- a/src/data/constants/index.js
+++ b/src/data/constants/index.js
@@ -33,6 +33,14 @@ const COURSE_URL_SLUG_VALIDATION_MESSAGE = {
   [COURSE_URL_SLUG_PATTERN]: 'Course URL slug contains lowercase letters, numbers, underscores, and dashes only and must be in the format <custom-url-slug> or learn/<primary_subject>/<org-slug>-<course_slug>.',
 };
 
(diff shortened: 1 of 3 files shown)


AfaqShuaib09 (approved):
Looks good to me, better to have a unittest for this usecase

zawan-ila:
> Looks good to me, better to have a unittest for this usecase

I don't see many tests for `discoveryApiService`. I could add a test that asserts that the include_restricted param is included in the url, but I am not sure how helpful that would be. If you think that would be good enough, I am happy to add it. Or if you have ideas for a different test, please let me know about that too.

src/data/services/DiscoveryDataApiService.js
```diff
@@ -10,6 +11,7 @@ class DiscoveryDataApiService {
     const queryParams = {
       editable: 1,
       exclude_utm: 1,
+      include_restricted: restrictionTypeValues,
```
DawoudSheraz, line 14: hmmm, do we not need any test updates for this?

src/data/constants/index.js
```diff
@@ -33,6 +33,14 @@ const COURSE_URL_SLUG_VALIDATION_MESSAGE = {
   [COURSE_URL_SLUG_PATTERN]: 'Course URL slug contains lowercase letters, numbers, underscores, and dashes only and must be in the format <custom-url-slug> or learn/<primary_subject>/<org-slug>-<course_slug>.',
 };
 
+const RESTRICTION_TYPE_OPTIONS = [
+  { value: '', label: '--------' },
+  { value: 'custom-b2b-enterprise', label: 'Custom Enterprise' },
+  { value: 'custom-b2c', label: 'Custom B2C' },
+];
+
+const RESTRICTION_TYPE_VALUES = RESTRICTION_TYPE_OPTIONS.map(opt => opt.value).filter(Boolean).join(',');
```
DawoudSheraz, line 42: 👍🏽 on that Boolean filter

DawoudSheraz (approved):

src/data/services/DiscoveryDataApiService.js
zawan-ila: No. There are virtually no tests for this file.

PR #966, opened 2024-08-02
fix: add default value when destructuring collaborator options

zawan-ila:
## Overview

It can happen that the collaborators have not been fetched before the page renders. This causes the collaborators to evaluate to undefined raising TypeErrors later when we try to use a `.map` on them

### DIFF
--- a/src/components/EditCoursePage/EditCourseForm.jsx
+++ b/src/components/EditCoursePage/EditCourseForm.jsx
@@ -268,7 +268,7 @@ export class BaseEditCourseForm extends React.Component {
 
     const {
       data: {
-        results: allResults,
+        results: allResults = [],
       },
     } = collaboratorOptions;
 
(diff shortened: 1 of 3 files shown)


AfaqShuaib09 (approved):

src/components/EditCoursePage/EditCourseForm.jsx
```diff
@@ -268,7 +268,7 @@ export class BaseEditCourseForm extends React.Component {
 
     const {
       data: {
-        results: allResults,
+        results: allResults = [],
```
DawoudSheraz, line 271: if the collaborators are not fetched, how will the retry occur if data is set to empty list here?

zawan-ila: Its just setting a default. Once the collaborators have been fetched, the state will update which will update the collaborator options with a "non-undefined" value for `allResults` and then that will be used by the code.

[PS] Note that the collaborator info is in collaboratorOptions, and this is just pulling information out of that object.

PR #990, opened 2024-10-10
fix: update default time from 12:00 to 00:00

hamza-56:
### [PROD-4200](https://2u-internal.atlassian.net/browse/PROD-4200)

### DIFF
--- a/src/components/DateTimeField/index.jsx
+++ b/src/components/DateTimeField/index.jsx
@@ -21,7 +21,7 @@ class DateTimeField extends React.Component {
         : getDateWithDashes(this.props.input.value),
       time: (this.props.utcTimeZone
         ? getTimeStringUTC(this.props.input.value)
-        : getTimeString(this.props.input.value)) || '12:00',
+        : getTimeString(this.props.input.value)) || '00:00',
     };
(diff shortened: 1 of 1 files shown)


Ali-D-Akbar (approved):

src/components/DateTimeField/index.jsx
```diff
@@ -21,7 +21,7 @@ class DateTimeField extends React.Component {
         : getDateWithDashes(this.props.input.value),
       time: (this.props.utcTimeZone
         ? getTimeStringUTC(this.props.input.value)
-        : getTimeString(this.props.input.value)) || '12:00',
+        : getTimeString(this.props.input.value)) || '12:00 AM',
```
DawoudSheraz, line 24: Does this AM work out of the box? Should it not be 00:00?

DawoudSheraz (commented):
The commit type should be chore, not fix.

src/components/DateTimeField/index.jsx
hamza-56: Screenshot 2024-10-11 at 3 12 19 PM
Yes, you're right.
Updated

DawoudSheraz (approved):

PR #991, opened 2024-10-30
fix: tinyMCELanguageSelectorPlugin error related to Button ID

AfaqShuaib09:
This PR fixes the TinyMCE editor Language Selector `failed to find and set button ID in TinyMceLanguageSelectorPlugin after 3 attempts`. It adds a `0.1-sec` timeout to prevent TinyMCE from appearing immediately after the state change, as the editor reloads afterward, causing it to disappear and triggering the error above.

Error Reproduction Steps:

- Pull the code from the master branch.
- Create a new course run for any existing course.
- Add the relevant details and move the course to the review state. You should see the message `failed to find and set button ID in TinyMceLanguageSelectorPlugin after 3 attempts in the console log`.
- Checkout to this branch.
- Create a new course run for any existing course.
- You will no longer see this error in the console logs.

### DIFF
--- a/src/components/EditCoursePage/CollapsibleCourseRun.jsx
+++ b/src/components/EditCoursePage/CollapsibleCourseRun.jsx
@@ -4,7 +4,6 @@ import React from 'react';
 import { CopyToClipboard } from 'react-copy-to-clipboard';
 import moment from 'moment';
 import PropTypes from 'prop-types';
-import { debounce } from 'lodash';
 
 import { Field, FieldArray } from 'redux-form';
 import { getAuthenticatedUser } from '@edx/frontend-platform/auth';
--- a/src/components/EditCoursePage/CollapsibleCourseRun.jsx
+++ b/src/components/EditCoursePage/CollapsibleCourseRun.jsx
@@ -281,7 +281,7 @@ class CollapsibleCourseRun extends React.Component {
 
     if (courseInReview !== isCourseRunInReview) {
         setTimeout(() => {
-            this.setState({ isCourseRunInReview: courseInReview });
+            this.setState({ isCourseRunInReview: IN_REVIEW_STATUS.includes(courseRun.status) });
         }, 500);
     }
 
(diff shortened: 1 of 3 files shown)


src/components/EditCoursePage/CollapsibleCourseRun.jsx
```diff
@@ -278,6 +279,12 @@ class CollapsibleCourseRun extends React.Component {
     const productSource = courseInfo?.data?.product_source?.slug;
     const courseType = courseInfo?.data?.course_type;
 
+    if (IN_REVIEW_STATUS.includes(courseRun.status) !== isCourseRunInReview) {
+      setTimeout(() => {
+        this.setState({ isCourseRunInReview: IN_REVIEW_STATUS.includes(courseRun.status) });
+      }, 100);
```
DawoudSheraz, line 285: this 0.1 value -- what was the reason for selecting this value?

AfaqShuaib09: I tried random values for the timeout, `i.e.` 500, 200, 100, 50. It's minimum value at which this error no longer appears.

DawoudSheraz: That is from your local testing, I would suggest keeping the deployment environment in mind when deciding this.

zawan-ila (commented):
Can you please add instructions to reproduce the issue locally in the PR description? Some details are present on our internal ticket, but would be great to add them here as well.

AfaqShuaib09:
> Can you please add instructions to reproduce the issue locally in the PR description? Some details are present on our internal ticket, but would be great to add them here as well.

Done, I've added the steps to reproduce this error locally, you can test it out on stage as well.

DawoudSheraz (approved):

zawan-ila (commented):
I can repro the error by simply opening a course. And checkout out this code does not seem to fix it. Could it be that my local data is simply messed up?

AfaqShuaib09:
> failed to find and set button ID in TinyMceLanguageSelectorPlugin after 3 attempts

Ah, it’s reproducible for me now, even after reloading the page. However it’s sort of flaky behavior, sometimes it appears on page refresh, and sometimes it doesn’t. Two days ago, it didn’t appear for me at all after refreshing the page. I’m not sure what the issue is. The PR above will resolve the error for the review state transition scenario. I think it would be better to ignore this error, as we did for support-tools

PR #992, opened 2024-11-12
chore: add resize option to adjust the height of tinymce editor manually

AfaqShuaib09:
[PROD-3235](https://2u-internal.atlassian.net/browse/PROD-3235)
-------
This PR adds vertical resize option to tinymce editor to adjust the size manually upon dragging from right bottom.
https://www.tiny.cloud/docs/tinymce/latest/editor-size-options/#controlling-user-editor-resizing.

https://github.com/user-attachments/assets/e6b83811-3a8d-41f2-81de-16da2fc08a03


Ali-D-Akbar (approved):

src/components/RichEditor/index.jsx
```diff
@@ -89,7 +90,9 @@ class RichEditor extends React.Component {
             init={{
               branding: false,
               menubar: false,
-              plugins: 'legacyoutput link lists language directionality',
+              plugins: 'legacyoutput link lists language directionality autoresize',
+              min_height: 200,
+              max_height: 400,
```
DawoudSheraz, line 95: I am not sure if this is intended behavior. While it is good to auto resize, the user should be given control to change the size as well (at least that's what appears to be actual ask in the ticket)

AfaqShuaib09: I've looked into the TinyMCE docs, and there isn't any built-in plugin for adjustable-height. However, we can create a custom plugin for adjustable height.
https://www.tiny.cloud/docs/tinymce/latest/creating-a-plugin/

The reporter in the ticket also mentioned that the older version had resizable fields, so I strongly suspect it's about autoresize rather than manually adjusting the height. Plus it seems like not a good user experience to require users to adjust the editor size themselves.

AfaqShuaib09: Created a custom height-adjustable plugin locally, and this is how it looks. Thoughts on this? 
I think the auto-resize plugin will be better here, as users don’t have to adjust the height of the editor’s text area manually.


https://github.com/user-attachments/assets/33e071bf-d96c-44ac-9118-365d6fc3d77c

DawoudSheraz: This is not right. I think if we can use drag-and-drop based resize (like in textarea), that would be good. Otherwise auto-resize option is going to be there.

AfaqShuaib09: I checked into the `drag-and-drop` tinymce plugin, but it’s more focused on adding images/files to the text editor rather than adjusting the height. 
https://www.tiny.cloud/blog/drag-and-drop/.

I think `auto-resize` is good for now.

DawoudSheraz (approved):

Ali-D-Akbar (approved):

PR #1000, opened 2025-02-20
feat: make level type and subjects field optional for masters courses

Ali-D-Akbar:
[PROD-3254](https://2u-internal.atlassian.net/browse/PROD-3254)
This PR puts Level Type and Primary Subject as non-marketable fields i.e. make them optional for masters courses (and all the non-marketable courses)


DawoudSheraz:
Instead of hiding the fields altogether, make them optional in case some teams want to add the values.

DawoudSheraz (approved):

AfaqShuaib09 (approved):

PR #1009, opened 2025-04-29
feat!: update tinymce-react to v6.1.0

DawoudSheraz:
### Description
The PR updates tinymce-react to v6.1.0. In doing this change, there have been some changes in how the editor content state is managed. 

- Instead of using initialValue, [value](https://www.tiny.cloud/docs/tinymce/latest/react-ref/#value) prop is used to manage the editor content. While both initial value and value can be used together, it did not cause any impact on the functionality of the editor. 
- Make component controlled. Previously, the component was using [OnChange](https://www.tiny.cloud/docs/tinymce/latest/react-ref/#event-binding) to handle character count change and propagate content to redux form field. This did not work correctly when initialValue was changed to value. The editor state needed to be changed. For this purpose, alongside OnChange, [OnEditorChange](https://www.tiny.cloud/docs/tinymce/latest/react-ref/#oneditorchange) prop has been added. This prop is tinymce-react specific and should be used when using editor as controlled.

### DIFF
--- a/src/components/RichEditor/index.jsx
+++ b/src/components/RichEditor/index.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-
 import { Editor } from '@tinymce/tinymce-react';
 import 'tinymce/tinymce.min';
 import 'tinymce/icons/default';
@@ -21,9 +20,20 @@ class RichEditor extends React.Component {
     super(props);
     this.state = {
       charCount: 0,
+      value: props.input.value,
     };
     this.initCharCount = this.initCharCount.bind(this);
     this.updateCharCount = this.updateCharCount.bind(this);
+    this.handleEditorChange = this.handleEditorChange.bind(this);
+    this.handleOnChange = this.handleOnChange.bind(this);
+  }
+
+  handleEditorChange(newValue, editor) {
+    this.updateCharCount(editor);
+  }
+
+  handleOnChange(event, editor) {
+    this.updateCharCount(editor);
   }
 
   initCharCount(event, editor) {
@@ -33,12 +43,16 @@ class RichEditor extends React.Component {
     });
   }
 
-  updateCharCount(event, editor) {
+  updateCharCount(editor) {
     const content = editor.getContent({ format: 'text' });
+    const htmlContent = editor.getContent();
+
     this.setState({
       charCount: content.length,
+      value: htmlContent,
     });
-    const htmlContent = editor.getContent();
+
+    // propagate the changes to redux form field
     this.props.input.onChange(htmlContent);
   }
 
@@ -48,7 +62,6 @@ class RichEditor extends React.Component {
       id,
       label,
       input: {
-        value,
         name,
       },
       meta: {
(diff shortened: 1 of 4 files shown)


src/components/RichEditor/index.jsx
```diff
@@ -103,8 +128,9 @@ class RichEditor extends React.Component {
               content_style: contentStyle,
               default_link_target: '_blank',
             }}
-            onChange={this.updateCharCount}
-            onKeyUp={this.updateCharCount}
+            onChange={this.handleOnChange}
+            onEditorChange={this.handleEditorChange}
```
AfaqShuaib09, line 132: [curious] Why has this been added here?

DawoudSheraz: It is already explained in PR description. To summarize, onEditorChange is tinymce's own event for detecting content changes and is recommended to be used, especially when making the editor a controlled component.

src/components/RichEditor/index.jsx
```diff
@@ -33,12 +55,16 @@ class RichEditor extends React.Component {
     });
   }
 
-  updateCharCount(event, editor) {
+  updateCharCount(editor) {
```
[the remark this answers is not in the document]
DawoudSheraz: No. This was called as onchange handler before. Instead of calling this method, I have added a new wrapper method (handleOnChange) that calls this method. The purpose of doing so was the introduction of handleEditorChange which performs the same operation but its first argument is editor, not event.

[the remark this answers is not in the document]
AfaqShuaib09: ok got it

Ali-D-Akbar (approved):
Tested locally. The changes are being updated correctly in all richtext fields.

AfaqShuaib09 (approved):

AfaqShuaib09 (approved):

src/components/RichEditor/index.jsx
```diff
@@ -21,9 +20,32 @@ class RichEditor extends React.Component {
...
+
+  componentDidUpdate(prevProps) {
+    const {
+      input: {
+        value,
+      },
+    } = this.props;
+
+    if (value !== prevProps.input.value) {
+      this.setState({ value });
```
zawan-ila, line 39: Is there no need to preserve the rest of the `state` (charCount) here?

src/components/RichEditor/index.jsx
zawan-ila, line 132: Can you provide a "noobs" explanation for when `onChange` and `onEditorChange` will exactly be triggered?

DawoudSheraz: OnEditorChange is for all things editor-related (content changes, copy/paste, undo/redo) while onChange is on the DOM level (focus, blur, etc.). TinyMCE renders as an iframe, and to have better control over how the content is managed, it has a separate event for Editor, to distinguish from DOM events.

src/components/RichEditor/index.jsx
DawoudSheraz: value state is set explicitly based on props (input.value), charCount is not. I did not observe any issue with charCount not being defined here.

zawan-ila (approved):

PR #1012, opened 2025-05-12
feat: Bulk Operation Task Details Page

AfaqShuaib09:
[PROD-4368](https://2u-internal.atlassian.net/browse/PROD-4368)
-----------
This PR adds the `Bulk Operation Task Details` page, which displays detailed attributes for a single record

image

https://github.com/user-attachments/assets/d06cba96-1826-4974-90c2-0e366d9495ad

image

### DIFF
--- a/src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
+++ b/src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
@@ -1,0 +1,196 @@
+import React, { useState, useMemo } from 'react';
+import 'moment-timezone';
+import PropTypes from 'prop-types';
+import Papa from 'papaparse';
+import {
+  Button, DataTable, useToggle, FullscreenModal, Stack,
+} from '@openedx/paragon';
+
+import { formatDateTime } from '../../utils';
+import './styles.scss';
+
+const ImageCell = ({ value }) => (
+  value ? (
+    <a href={value} target="_blank" rel="noopener noreferrer">
+      View Image
+    </a>
+  ) : null
+);
+
+const parseCSV = async (url) => {
+  const response = await fetch(url);
+  const text = await response.text();
+
+  return new Promise((resolve, reject) => {
+    Papa.parse(text, {
+      header: true,
+      skipEmptyLines: true,
+      complete: (results) => {
+        const { data, meta } = results;
+        const headers = meta.fields;
+        resolve({ headers, rows: data });
+      },
+      error: (error) => {
+        reject(error);
+      },
+    });
+  });
+};
+
+const BulkOperationDetails = ({ task }) => {
+  const [csvRows, setCsvRows] = useState([]);
+  const [csvHeaders, setCsvHeaders] = useState([]);
+  const [isOpen, open, close] = useToggle(false);
+  const [csvPreviewError, setCsvPreviewError] = useState(null);
+  const DEFAULT_CSV_PREVIEW_PAGE_SIZE = 10;
+
+  const taskInfoRows = [
+    { key: 'Task ID', value: task.task_id },
+    { key: 'Task type', value: task.task_type },
+    { key: 'Uploaded by', value: task.uploaded_by },
+    { key: 'Status', value: task.status },
+    { key: 'Created', value: formatDateTime(task.created) },
+    { key: 'Modified', value: formatDateTime(task.modified) },
+  ];
+
+  const taskInfoColumns = [
+    { Header: 'Field', accessor: 'key' },
+    { Header: 'Value', accessor: 'value' },
+  ];
+
+  const handlePreviewCSV = async () => {
+    try {
+      const { headers, rows } = await parseCSV(task.csv_file);
+      setCsvHeaders(headers);
+      setCsvRows(rows);
+    } catch (error) {
+      console.error('Failed to preview CSV:', error);
+      setCsvPreviewError(`Failed to preview CSV: ${error.message}`);
+    }
+  };
+
+  const getDataTableColumns = (headers) => headers.map((header) => {
+    const isImageColumn = header.toLowerCase() === 'image';
+
+    const column = {
+      Header: header,
+      accessor: header,
+    };
+
+    if (isImageColumn) {
+      column.Cell = ImageCell;
+    }
+
+    return column;
+  });
+
+  const columns = useMemo(() => { // eslint-disable-line react-hooks/rules-of-hooks
+    if (!csvHeaders || csvHeaders.length === 0) {
+      return [];
+    }
+    return getDataTableColumns(csvHeaders);
+  }, [csvHeaders]);
+
+  const data = useMemo(() => { // eslint-disable-line react-hooks/rules-of-hooks
+    if (!csvHeaders || csvHeaders.length === 0 || !csvRows || csvRows.length === 0) {
+      return [];
+    }
+    return csvRows;
+  }, [csvRows, csvHeaders]);
+
+  return (
+    <div className="container mt-2">
+      <h2 className="mb-4">Task Details</h2>
+      <div className="bulk-op-details-table-container">
+        <DataTable
+          columns={taskInfoColumns}
+          data={taskInfoRows}
+          showPagination={false}
+        />
+      </div>
+
+      <Stack direction="horizontal" gap={3} className="mb-4 mt-4">
+        <a
+          href={task.csv_file}
+          className="btn btn-outline-primary"
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          Download CSV
+        </a>
+        <Button
+          variant="primary"
+          onClick={() => {
+            handlePreviewCSV();
+            open();
+          }}
+        >
+          Preview CSV
+        </Button>
+        <FullscreenModal
+          title="CSV Preview"
+          isOpen={isOpen}
+          onClose={close}
+          isOverflowVisible={false}
+        >
+          <div className="csv-preview-modal">
+            {csvPreviewError && (
+              <div className="alert alert-danger" role="alert">
+                {csvPreviewError}
+              </div>
+            )}
+            <DataTable
+              className="bulk-op-table"
+              columns={columns}
+              data={data}
+              isPaginated
+              initialState={{
+                pageSize: DEFAULT_CSV_PREVIEW_PAGE_SIZE,
+              }}
+              itemCount={csvRows.length}
+              data-testid="csv-table"
+            />
+            <DataTable.EmptyTable content="No results found" />
+            <DataTable.TableFooter />
+          </div>
+        </FullscreenModal>
+      </Stack>
+
+      {task.task_summary && (
+        <div className="bg-light p-3 rounded">
+          <h4>Task Summary</h4>
+          <pre className="bg-white p-2 rounded border">
+            {JSON.stringify(task.task_summary, null, 2)}
+          </pre>
+        </div>
+      )}
+    </div>
+  );
+};
+
+ImageCell.propTypes = {
+  value: PropTypes.string,
+};
+
+ImageCell.defaultProps = {
+  value: null,
+};
+
+BulkOperationDetails.propTypes = {
+  task: PropTypes.shape({
+    task_id: PropTypes.string.isRequired,
+    task_type: PropTypes.string.isRequired,
+    uploaded_by: PropTypes.string.isRequired,
+    status: PropTypes.string.isRequired,
+    created: PropTypes.string.isRequired,
+    modified: PropTypes.string.isRequired,
+    csv_file: PropTypes.string.isRequired,
+    task_summary: PropTypes.object, // eslint-disable-line react/forbid-prop-types
+  }),
+};
+
+BulkOperationDetails.defaultProps = {
+  task: null,
+};
+
+export default BulkOperationDetails;
--- a/src/components/BulkOperationDetailsPage/BulkOperationDetails.test.jsx
+++ b/src/components/BulkOperationDetailsPage/BulkOperationDetails.test.jsx
@@ -1,0 +1,117 @@
+import React from 'react';
+import '@testing-library/jest-dom';
+import {
+  render, screen, fireEvent, waitFor,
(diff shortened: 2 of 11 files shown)


src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
```diff
@@ -0,0 +1,155 @@
...
+ImageCell.defaultProps = {
+  value: null,
+};
+
+const parseCSV = async (url) => {
+  const response = await fetch(url);
+  const text = await response.text();
+  const lines = text.trim().split('\n');
+
+  const headers = lines[0].split(/,(?=(?:(?:[^"]*"){2})*[^"]*$)/);
```
DawoudSheraz, line 32: ah, what's the reason for using this regex?

src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
```diff
@@ -0,0 +1,155 @@
...
+    return <p className="text-info"> No BulkOperationTask found against the given ID. Please check the ID and try again.</p>;
+  }
+
+  const handlePreviewCSV = async () => {
+    try {
+      const { headers, rows } = await parseCSV(task.csv_file);
+      setCsvHeaders(headers);
+      setCsvRows(rows);
+    } catch (error) {
+      console.error('Failed to preview CSV:', error);
```
DawoudSheraz, line 56: The error would need to be displayed to the user to show that the preview has gone wrong.

src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
```diff
@@ -0,0 +1,155 @@
...
+
+  return (
+    <div className="container mt-2">
+      <h2 className="mb-4">Task Details</h2>
+      <p><strong>Task ID:</strong> {task.task_id}</p>
+      <p><strong>Task type:</strong> {task.task_type}</p>
+      <p><strong>Uploaded by:</strong> {task.uploaded_by}</p>
+      <p><strong>Status:</strong> {task.status}</p>
+      <p><strong>Created:</strong> {task && formatDateTime(task.created)}</p>
+      <p><strong>Modified:</strong> {task && formatDateTime(task.modified)}</p>
```
DawoudSheraz, line 83: nit: you can use table instead.

src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
```diff
@@ -0,0 +1,155 @@
...
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          Download CSV
+        </a>
+        <Button
+          className="ms-2"
+          variant="primary"
+          onClick={() => {
+            handlePreviewCSV();
```
DawoudSheraz, line 98: dont we need to await here as handlePreviewCSV is async?

src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
```diff
@@ -0,0 +1,155 @@
...
+        <FullscreenModal
+          title="CSV Preview"
+          isOpen={isOpen}
+          onClose={close}
+          isOverflowVisible={false}
+        >
+          <DataTable
+            className="bulk-op-table"
+            columns={getDataTableColumns(csvHeaders)}
+            data={csvRows.map((row) => row.reduce((acc, value, index) => {
```
DawoudSheraz, line 113: nit: acc --> accumulator

src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
AfaqShuaib09: To split a CSV line correctly on commas, only those commas that are not inside double quotes, I added this regex to fix issues I was having with columns being split incorrectly.

src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
AfaqShuaib09: I like the current imp better than a table. However, I'll update the code and add a screenshot here, then we can decide.

AfaqShuaib09: image

DawoudSheraz: Well, this looks better.

Ali-D-Akbar (commented):
Bunch of suggestions and questions.

src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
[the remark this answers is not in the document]
AfaqShuaib09: Yes, I just added a line below it to show an error to the user in such a case.

src/components/BulkOperationDetailsPage/styles.scss
```diff
@@ -0,0 +1,23 @@
...
+        th {
+            border: 1px solid black;
+            min-width: $cell-widths;
+            max-width: $cell-widths;
+        }
+        td {
+            border: 1px solid black;
+            min-width: $cell-widths;
+            max-width: $cell-widths;
+            text-overflow: ellipsis !important;
```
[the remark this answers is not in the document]
AfaqShuaib09: Yeah, just removed it 😄

src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
```diff
@@ -0,0 +1,220 @@
...
+          columns={taskInfoColumns}
+          data={taskInfoRows}
+          showPagination={false}
+        />
+      </div>
+
+      <Stack direction="horizontal" gap={3} className="mb-4 mt-4">
+        <a
+          href={task.csv_file}
+          className="btn btn-outline-primary me-2"
```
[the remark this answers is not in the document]
AfaqShuaib09: Ah, previously I added margins to create spacing between the buttons. I've already updated the code to add the spacing by using an outer stack container but forgot to remove the class from here.

src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
```diff
@@ -0,0 +1,220 @@
...
+    { Header: 'Field', accessor: 'key' },
+    { Header: 'Value', accessor: 'value' },
+  ];
+
+  useEffect(() => {
+    const debounceDelay = 500;
+
+    const timeout = setTimeout(() => {
+      if (!task || Object.keys(task).length === 0) {
+        setShowNotFound(true);
```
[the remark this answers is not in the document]
AfaqShuaib09: Yeah, this debounce effect is no longer required here. I've already handled it in index.js using the loading spinner update. Thanks for pointing it out.

zawan-ila (approved):

src/components/BulkOperationDetailsPage/BulkOperationDetails.test.jsx
```diff
@@ -0,0 +1,131 @@
...
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('renders task details correctly', () => {
+    render(
+      <IntlProvider locale="en">
+        <BulkOperationDetails task={mockTask} />
+      </IntlProvider>,
+    );
```
Ali-D-Akbar, line 52: Wrap this repeated render logic in a helper

src/components/BulkOperationDetailsPage/styles.scss
```diff
@@ -0,0 +1,37 @@
...
+
+.bulk-op-details-table-container {
+  .pgn__data-table-status-bar {
+    display: none;
+  }
+
+  .pgn__data-table-footer {
+    display: none;
+   }
+}
```
Ali-D-Akbar, line 37: nit: missing end of line

src/components/BulkOperationDetailsPage/styles.scss
```diff
@@ -0,0 +1,37 @@
+.csv-preview-modal {
+
+    $cell-widths: 220px;
```
Ali-D-Akbar, line 3: nit: Conventionally, this variable should be at the top... Outside `.csv-preview-modal`

src/data/services/DiscoveryDataApiService.test.js
```diff
@@ -20,3 +31,32 @@ describe('fetchOrganizationUsers', () => {
...
+
+    const result = await DiscoveryDataApiService.fetchBulkOperationTask(taskId);
+    expect(get).toHaveBeenCalledWith(expectedUrl);
+    expect(result.data).toEqual(mockData);
+  });
+
+  it('should return null on 404 response', async () => {
+    get.mockRejectedValue({ response: { status: 404 } });
+
+    const result = await DiscoveryDataApiService.fetchBulkOperationTask(taskId).catch(e => e);
```
Ali-D-Akbar, line 52: Do we usually use catch with await? I feel like we can use `.rejects` and expect a value in case a promise is rejected. Something like

`await expect(promise).rejects.toEqual(expectedError);`

Ali-D-Akbar (commented):
Looks good to me. Going to test this locally.

src/components/BulkOperationDetailsPage/BulkOperationDetails.test.jsx
AfaqShuaib09: Moved rendering to `beforeEach` method

src/data/services/DiscoveryDataApiService.test.js
AfaqShuaib09: Actually, I think both are fine. It's about preference of what you'd like. However, I've updated the code as you suggested.

src/utils/index.js
```diff
@@ -21,6 +21,7 @@ const getFormattedUTCTimeString = date => (date ? moment.utc(date).format('h:mm
 const localTimeZone = moment.tz(moment.tz.guess()).zoneAbbr();
 const utcTimeZone = moment.utc().zoneAbbr();
 const formatDate = date => (date ? moment(date).format('MMM DD, YYYY') : '');
+const formatDateTime = (dateStr) => (dateStr ? moment.tz(dateStr, 'UTC').format('MMM DD, YYYY, hh:mm:ss A') : '');
```
DawoudSheraz, line 24: nit: missing unit tests for this utility.

src/components/BulkOperationDetailsPage/BulkOperationDetails.test.jsx
```diff
@@ -0,0 +1,111 @@
...
+  });
+
+  it('renders BulkOperationDetails and previews CSV', async () => {
+    const previewButton = screen.getByRole('button', { name: /Preview CSV/i });
+    fireEvent.click(previewButton);
+
+    await waitFor(() => {
+      expect(screen.getByRole('dialog')).toBeInTheDocument();
+      const table = screen.getByRole('table');
+      expect(table).toBeInTheDocument();
```
DawoudSheraz, line 100: nit: should we not assert on parts of content?

src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
```diff
@@ -0,0 +1,196 @@
...
+    };
+
+    if (isImageColumn) {
+      column.Cell = ImageCell;
+    }
+
+    return column;
+  });
+
+  const columns = useMemo(() => { // eslint-disable-line react-hooks/rules-of-hooks
```
DawoudSheraz, line 87: Why is eslint disabled?

src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
```diff
@@ -0,0 +1,196 @@
...
+    }
+    return getDataTableColumns(csvHeaders);
+  }, [csvHeaders]);
+
+  const data = useMemo(() => { // eslint-disable-line react-hooks/rules-of-hooks
+    if (!csvHeaders || csvHeaders.length === 0 || !csvRows || csvRows.length === 0) {
+      return [];
+    }
+    return csvRows;
+  }, [csvRows, csvHeaders]);
```
DawoudSheraz, line 99: why is useMemo checking on both rows and headers?

src/components/BulkOperationDetailsPage/BulkOperationDetails.test.jsx
AfaqShuaib09: Sure we can, but there i a sort of dummy data with 2 fields, therefore i didn't add it in the test

src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
AfaqShuaib09: It's possibly a false positive from es-lint. The changes comply with all the rules mentioned here https://legacy.reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level.

src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
AfaqShuaib09: Shouldn't we refresh data, either of them changes?
I think there’s a possibility that the CSV might contain only headers, just thinking it as a case.

DawoudSheraz: Then should this not be done for headers useMemo too?

src/components/BulkOperationDetailsPage/BulkOperationDetails.test.jsx
DawoudSheraz: Let's add the assertions, since the code is previewing CSV and does not care about its structure. The test will verify that preview is working as expected.

src/components/BulkOperationDetailsPage/BulkOperationDetails.jsx
AfaqShuaib09: hmm yes i believe.

Ali-D-Akbar (approved):
I have tested the code locally. Looks good to me overall.

PR #1028, opened 2025-05-31
fix: fix skipped tests in ListField component

AfaqShuaib09:


DawoudSheraz (approved):
Commit type should be test, not fix.

src/components/ListField/ListField.test.jsx
```diff
@@ -176,185 +167,116 @@ describe('ListField - Collaborators', () => {
...
     const input = screen.getByRole('textbox');
     await userEvent.type(input, 'mit');
     const suggestionsList = await screen.findAllByTestId('list-field-suggestion');
-    waitFor(() => expect(suggestionsList[0]).toHaveAttribute('name', 'MIT'));
-    waitFor(() => expect(suggestionsList[0]).toHaveAttribute('uuid', 'a7d0e2c0-9a02-421b-93bf-d081339090cc'));
-    waitFor(() => expect(suggestionsList[0]).toHaveAttribute('item_text', 'Add New Collaborator'));
-    waitFor(() => expect(suggestionsList[0].url).not.toBeNull());
-    // check that clearing suggestions...clears suggestions
+    expect(within(suggestionsList[0]).getByAltText(/logo for mit/i)).toBeInTheDocument();
+    expect(within(suggestionsList[0]).getByText('MIT')).toBeInTheDocument();
```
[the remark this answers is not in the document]
AfaqShuaib09: uuid isn't available here as an attribute or visible on the screen, though we can extract it from the image url since it's part of the url.

src/components/ListField/ListField.test.jsx
```diff
@@ -176,185 +167,116 @@ describe('ListField - Collaborators', () => {
...
-    };
-    // Pretend we dragged the first item outside the list.
-    component.instance().onDragEnd(result);
-
-    // Verify that the onChange method has NOT been called
-    expect(collaboratorInput.onChange).not.toBeCalled();
-    expect(firstItem).toEqual(component.state().currentList[0].uuid);
+  it('gets no suggestions for short autocomplete', async () => {
+    render(<MemoryRouter><ListField {...collaboratorDefaultProps} /></MemoryRouter>);
+    const input = await screen.getByRole('textbox');
```
hamza-56, line 237: nit: I noticed we're using `await` with `screen.getBy*` in some places, but not in others. Since getBy* is a synchronous method and doesn't require `await`, could we remove it where it's not needed and make the usage consistent throughout the file?

AfaqShuaib09: Ah, I was facing issues with these two tests, so I tried to fix them using await. In the end, it turned out to be a different issue, which I resolved. But I forgot to remove these calls. Thanks for pointing it out.
Lemme remove them from these tests real quick

hamza-56 (approved):

PR #1029, opened 2025-06-02
Add await to waitFor calls in tests

hamza-56:
### [PROD-4381](https://2u-internal.atlassian.net/browse/PROD-4381)

In React 18 upgrade PR, the tests were moved to RTL and many tests have been using waitFor and screen.findBy. They need proper async/await to ensure the tests work as expected.


DawoudSheraz (approved):

src/components/SidePanes/User.test.jsx
```diff
@@ -1,24 +1,22 @@
...
 describe('User', () => {
-  it('shows a remove button iff onRemove is set', () => {
-    const baseArguments = {
-      name: 'User',
-      userId: 0,
-    };
+  const baseArguments = {
+    name: 'User',
+    userId: 0,
+  };
```
AfaqShuaib09, line 11: [curious] why we have moved baseArguments out from the testcase here?

AfaqShuaib09: Ok, got my answer, we're using it in a new testcase

PR #1037, opened 2025-06-18
feat!: add design tokens support

brian-smith-tcril:
BREAKING CHANGE: Pre-design-tokens theming is no longer supported. 

# LAND AS SQUASH

The `master-design-tokens` branches were created as long running feature branches to ensure design tokens land on all MFEs simultaneously. These have been kept up to date via merge commits and should land on their respective MFE `master` branches as a single squashed commit.


.env.test
```diff
@@ -9,3 +9,4 @@ ADDITIONAL_METADATA_REQUIRED_FIELDS='{}'
 IS_NEW_SLUG_FORMAT_ENABLED='false'
 MARKETING_SITE_PREVIEW_URL_ROOT=''
 COURSE_URL_SLUGS_PATTERN = `{}`;
+PARAGON_THEME_URLS=
```
MaxFrank13, line 12:
```suggestion
PARAGON_THEME_URLS={}
```

package.json
```diff
@@ -68,7 +68,7 @@
   "devDependencies": {
     "@edx/browserslist-config": "^1.2.0",
     "@edx/stylelint-config-edx": "^2.3.0",
-    "@openedx/frontend-build": "^14.0.14",
+    "@openedx/frontend-build": "^14.3.1",
```
adamstankiewicz, line 71: [curious, non-blocking] Do we want to bump to latest version here?

